PR #62, opened 2024-07-09
feature/enable-ocpi-validation-for-koa-server-responses

elliot-sabitov:
feat: created new ValidatedResponseSchema annotation which reused ResponseSchema annotation directly but also sets the class into reflect metadata and then created ResponseValidationMiddleware which gets the class and annotation and builds the DTO via plainToClass and runs validation on the response that is going to be sent in response

To use, all we have to do is switch occurrences of @ResponseSchema to use @ValidatedResponseSchema which will enable the ResponseValidationMiddleware.

Screenshot 2024-07-09 at 3 57 24 PM
Screenshot 2024-07-09 at 3 57 10 PM

### DIFF
--- a/00_Base/src/model/ocpi.response.ts
+++ b/00_Base/src/model/ocpi.response.ts
@@ -19,10 +19,17 @@ export enum OcpiResponseStatusCode {
   HubConnectionProblem = 4003,
 }
 
+export enum OcpiResponseProps {
+  status_code = 'status_code',
+  status_message = 'status_message',
+  timestamp = 'timestamp',
+  data = 'data',
+}
+
 export class OcpiResponse<T> {
   @Enum(OcpiResponseStatusCode, 'OcpiResponseStatusCode')
   @IsNotEmpty()
-  status_code!: OcpiResponseStatusCode;
+  [OcpiResponseProps.status_code]!: OcpiResponseStatusCode;
   /**
    *
    * @type {string}
@@ -30,7 +37,7 @@ export class OcpiResponse<T> {
    */
   @IsString()
   @Optional()
-  status_message?: string;
+  [OcpiResponseProps.status_message]?: string;
 
   /**
    *
@@ -38,15 +45,14 @@ export class OcpiResponse<T> {
    * @memberof OcpiResponseDTO
    */
   @IsDate()
-  timestamp!: Date;
+  [OcpiResponseProps.timestamp]!: Date;
 
   /**
    *
    * @type {object}
    * @memberof OcpiResponseDTO
    */
-  @Optional()
-  data?: T;
+  [OcpiResponseProps.data]?: T;
 }
 
 export const buildOcpiResponse = <T>(
--- a/00_Base/src/util/decorators/ValidatedResponseSchema.ts
+++ b/00_Base/src/util/decorators/ValidatedResponseSchema.ts
@@ -10,6 +10,15 @@ import { ResponseValidationMiddleware } from '../middleware/ResponseValidationMi
 
 export const validatedResponseParam = 'validatedResponseParam';
 
+/**
+ * ValidatedResponseSchema decorator wraps {@link ResponseSchema} and adds the input
+ * responseClass into Reflect.metadata and enables the {@link ResponseValidationMiddleware}
+ * which will get the responseClass via Reflect and perform class-validator validation
+ * on the response body prior to sending it back to the client.
+ * @param responseClass - response class
+ * @param options - {@link ResponseSchema} options
+ * @constructor
+ */
 export const ValidatedResponseSchema = (
   responseClass: Constructable<any>,
   options: {
--- a/00_Base/src/util/decorators/ValidatedResponseSchema.ts
+++ b/00_Base/src/util/decorators/ValidatedResponseSchema.ts
@@ -28,8 +28,8 @@ export const ValidatedResponseSchema = (
     isArray?: boolean;
     examples?: any;
   } = {},
-) => {
-  return function (object: any, methodName: string) {
+) =>
+  function (object: any, methodName: string) {
     // reuse original
     ResponseSchema(responseClass, options)(
       object as any,
@@ -48,4 +48,3 @@ export const ValidatedResponseSchema = (
     // apply middleware
     UseAfter(ResponseValidationMiddleware)(object, methodName);
   };
-};
--- a/00_Base/src/util/decorators/ValidatedResponseSchema.ts
+++ b/00_Base/src/util/decorators/ValidatedResponseSchema.ts
@@ -1,0 +1,42 @@
+// Copyright (c) 2023 S44, LLC
+// Copyright Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache 2.0
+
+import { Constructable } from 'typedi';
+import { ResponseSchema } from '../../openapi-spec-helper';
+import { UseAfter } from 'routing-controllers';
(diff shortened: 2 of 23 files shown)


00_Base/src/model/ocpi.response.ts
```diff
@@ -19,34 +19,40 @@ export enum OcpiResponseStatusCode {
...
   @IsDate()
-  timestamp!: Date;
+  [OcpiResponseProps.timestamp]!: Date;
 
   /**
    *
    * @type {object}
    * @memberof OcpiResponseDTO
    */
-  @Optional()
```
langiole, line 48: Why is optional removed?

00_Base/src/util/middleware/ResponseValidationMiddleware.ts
```diff
@@ -0,0 +1,90 @@
...
+              .join('');
+            throw new ResponseValidationError(errorMessages);
+          }
+        } else {
+          // todo handle OcpiErrorResponse and OcpiEmptyResponse
+        }
+      }
+    } else {
+      throw new InternalServerError(
+        'Could not find matching route? This should not happen',
```
langiole, line 57: Very minor, but I'd just update the punctuation to:
`Could not find matching route. This should not happen.`

00_Base/src/util/decorators/ValidatedResponseSchema.ts
```diff
@@ -0,0 +1,42 @@
...
+// SPDX-License-Identifier: Apache 2.0
+
+import { Constructable } from 'typedi';
+import { ResponseSchema } from '../../openapi-spec-helper';
+import { UseAfter } from 'routing-controllers';
+import { ResponseValidationMiddleware } from '../middleware/ResponseValidationMiddleware';
+
+export const validatedResponseParam = 'validatedResponseParam';
+
+export const ValidatedResponseSchema = (
```
langiole, line 13: I'm struggling to understand what's happening here.

Maybe a doc is in order?

00_Base/src/util/middleware/ResponseValidationMiddleware.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  private findRoute = (ctx: Context) => {
+    const storage = getMetadataArgsStorage();
+    const matchedRoute = ctx._matchedRoute;
+    if (!matchedRoute) return null;
+    const method = ctx.request?.req?.method?.toLowerCase();
+    if (!method) return null;
+    const controllerMetadata = storage.controllers.find((controller) => {
+      return matchedRoute.startsWith(`${OcpiRoutePrefix}${controller.route}`);
```
langiole, line 69: How does this work for paths with params?

00_Base/src/util/middleware/ResponseValidationMiddleware.ts
```diff
@@ -0,0 +1,90 @@
...
+import {
+  OcpiResponse,
+  OcpiResponseStatusCode,
+  OcpiRoutePrefix,
+} from '../../index';
+import { ResponseValidationError } from '../../exception/ResponseValidationError';
+
+@Middleware({ type: 'before' })
+@Service()
+export class ResponseValidationMiddleware implements KoaMiddlewareInterface {
```
langiole, line 21: Might be nice to have a one liner javadoc describing the purpose of this middleware.

langiole: I'm under the impression that it is to add validation to outgoing responses.

elliot-sabitov: added

00_Base/src/util/middleware/ResponseValidationMiddleware.ts
elliot-sabitov: Good question - so it is best to loo at what `getMetadataArgsStorage()` provides, but essentially in there, among other things, it has a list of `controllers` that look like this:
```
{
  "type": "json",
  "target: Constructible,
  "route": "/:versionId/locations"
}
```
and `actions` that looks like this: 
```
{
  "type": "get",
  "method": "getLocationById",
  "route": "/:location_id"
  "target: Constructible,
},
{
  "type": "get",
  "method": "getLocations"
  "target: Constructible,
}
```
you can see that actions may or may not have a path. The code in this PR gets a `matchingRoute` from the context and is then able to match the route in the controller plus the action to figure out which method is actually being called. We need to do this so that we can call :
```
const responseClass = Reflect.getMetadata(
        validatedResponseParam,
        target.prototype,
        method,
      );
```
to get the actual class that was passed into the `@ValidatedResponseSchema` annotation which would look like this:
```
@ValidatedResponseSchema(LocationResponse, {
    statusCode: HttpStatus.OK,
    description: 'Successful response',
    examples: {
      success: MOCK_LOCATION,
    },
  })
```

00_Base/src/util/decorators/ValidatedResponseSchema.ts
elliot-sabitov: added

00_Base/src/util/util.ts
```diff
@@ -15,12 +15,14 @@ export enum CountryCode {
 }
 
 export const plainToClass = <T>(constructor: Constructable<T>, plain: T): T =>
```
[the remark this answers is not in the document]
elliot-sabitov: it is possible to pass additional options to `plainToInstance` and `instanceToPlain` methods so centralizing via a helper method like this can help us apply such a config across all usages - otherwise you are correct, it isnt needed, and  currently we don't pass in a config, as we previously did, but I think it is still good just in case we do face the need since we already have it...

00_Base/src/model/ocpi.response.ts
elliot-sabitov: Good question - so it turns out with Reflect and inheritance, what I saw that if lets say we have our structure:
```
export class OcpiResponse<T> {
  @Optional()
  data?: T;
}
```
and then we have:
```
export class CdrResponse extends OcpiResponse<CdrDTO> {
  Optional(false)
  data!: CdrDTO[]
}
```

then unfortunately it does not behave as you would expect. The optional in the derived class doesnt override things as you would assume it would. I have faced this in a few places and found that it is best to annotate the derived classes and keep base classes in simple TS without annotations unless there is an annotation that will truly be shared and enabled by all derived classes....

Removing this Optional because it is simply unnecessary because our derived classes already do define whether data is optional or not, so our base class is acting more like an interface here. Does that make sense? I can try and clarify further

00_Base/src/util/middleware/global.exception.handler.ts
```diff
@@ -109,6 +110,15 @@ export class GlobalExceptionHandler implements KoaMiddlewareInterface {
               ),
             );
             break;
+          case ResponseValidationError.name:
+            context.status = HttpStatus.BAD_REQUEST;
```
ChrisWeissmann, line 114: Are you sure you want to return 400 here? OCPI states in `5 Status codes`:

> The transport layer ends after a message is correctly parsed into a (semantically unvalidated) JSON structure. When a message does not contain a valid JSON string, the HTTP error 400 - Bad request MUST be returned.
> If a request is syntactically valid JSON and addresses an existing resource, a HTTP error MUST NOT be returned. Those requests are supposed to have reached the OCPI layer.

Since the validation introduced here is not actually dependent on the client, but rather internal, HTTP 400 doesn't seem correct.

PR #68, opened 2024-07-10
Bugfix: Missing Headers in Async Push

lydiazcheng:
## Changes
1. add from/to country code and party id info in start and stop command use cases.
2. refactor `send` in `AsyncResponder` so that it can be used for both Charging Profile and Command module.

### 5. Del Charging Profile Screenshot 2024-07-11 at 4 10 38 PM Screenshot 2024-07-11 at 4 11 16 PM


00_Base/src/util/AsyncResponder.ts
```diff
@@ -19,39 +19,26 @@ export class AsyncResponder {
...
-  async sendChargingProfileResult(
+  async send(
     correlationId: string,
     data:
       | ActiveChargingProfileResult
       | ClearChargingProfileResult
-      | ChargingProfileResult,
+      | ChargingProfileResult
+      | CommandResult,
+    sessionId?: string,
```
langiole, line 29: We pass in sessionId to figure out which party id/country code headers to use?

If so, can we instead of update to accept these headers are params instead of sessionId?

We should move the session look up logic out of this method and into whoever is calling this method.

My reasoning is that we are bleeding OCPP logic into this class when we do not have to (e.g. the reference to transaction repo).

langiole: Not to mention, it looks like the callers of this method already look up transaction before calling this method so we it seems like we have a redundant look up.

00_Base/src/util/command.executor.ts
```diff
@@ -89,6 +91,7 @@ export class CommandExecutor {
     await this.responseUrlRepo.saveResponseUrl(
       correlationId,
       stopSession.response_url,
+      stopSession.session_id,
```
langiole, line 94: Example of redundant transaction look up.

A few lines above this call we fetched transaction. But then in this saveRepsonseUrl method we seem to do another transaction look up.

Why not just extract the party id and country code from the transaction above and pass through into the method here?

lydiazcheng: @langiole These feedback are very helpful. Thank you! I updated the code. Please let know if there is any further ideas.

00_Base/src/trigger/AsyncReceiverApi.ts
```diff
@@ -9,28 +9,22 @@ export class AsyncReceiverApi extends BaseClientApi {
...
-          additionalHeaders: this.getOcpiHeaders(params),
-        },
-        body,
-      );
-    } else {
-      return this.createRaw<OcpiResponse<void>>(url, body, {}).then(
-        (response) => response.result,
-      );
-    }
+    params.authorization = await this.getAuthToken(
```
langiole, line 14: I almost feel like we should have some sort of request interceptor that auto fills in the auth header given the ocpi params.

For the other push triggers (sessions, cdrs, locations, etc.) do they look up the auth token themselves similar to this?

00_Base/src/util/AsyncResponder.ts
```diff
@@ -5,70 +5,35 @@ import { ActiveChargingProfileResult } from '../model/ActiveChargingProfileResul
...
+      | ChargingProfileResult
+      | CommandResult,
+    params?: OcpiParams,
   ) {
     const responseUrlEntity =
       await this.responseUrlRepo.getResponseUrl(correlationId);
     if (responseUrlEntity) {
-      if (!responseUrlEntity.sessionId) {
-        throw new NotFoundError('Session Id not found');
+      params = params ?? responseUrlEntity.params;
```
langiole, line 30: Minor, but I wonder if it is cleaner to always save the OcpiParams to response url entity.

Then, you would never need to pass into this method.

This would require you to always set the ocpi params from commands executer.

langiole (approved):

00_Base/src/util/AsyncResponder.ts
lydiazcheng: > Minor, but I wonder if it is cleaner to always save the OcpiParams to response url entity.
> 
> Then, you would never need to pass into this method.
> 
> This would require you to always set the ocpi params from commands executer.

This param is set as an optional column in table because in use case like start session command, the transaction id is not yet generated when we store the response url.

langiole: Yes transaction is not yet generated, but we would still know the ocpi params as this is given in the start session url params.

We just dont pass it down from the api layer.

I am not sure if @ChrisWeissmann or @elliot-sabitov implemented something where we can easily extract the ocpi headers from anywhere in the code, or if we'd be required to pass it down.

That being said, just a minor thought I had. PR looks good.

00_Base/src/trigger/AsyncReceiverApi.ts
lydiazcheng: @langiole It is a very good point. I am also struggling on this question. 

Currently cdrs, sessions and locations always broadcast. You can find the related part in each broadcaster, e.g., [SessionBroadcaster](https://github.com/citrineos/citrineos-ocpi/blob/main/00_Base/src/broadcaster/session.broadcaster.ts) and it calls this method [broadcastToClients](https://github.com/citrineos/citrineos-ocpi/blob/b249e20de3ba9af8cdd57d1e12ec8c0c503d1e7c/00_Base/src/trigger/BaseClientApi.ts#L259) to get tokens and endpoints. So get token does not happens in any client api scripts.

Currently AsyncReceiverApi and PutChargingProfile are the only 2 places which send to a single target client and do this getAuthToken. 

I was thinking about create sender scripts for aync response and charging profile which are similar to broadcasters then we move getAuthToken into senders?

langiole: I think what is here is sufficient.

00_Base/src/util/AsyncResponder.ts
elliot-sabitov: yep, if this is a Koa endpoint, you should be able to get it like this: ```@FunctionalEndpointParams() ocpiHeaders?: OcpiHeaders,``` in the request handler method

lydiazcheng: If we change to use `ocpiHeaders`, we would need to pass headers from request handler down to service and then command executor. I think we can keep the existing change for now and decide whether refactor to use `ocpiHeaders` for Command and Charging Profile apis later.

langiole: I agree @lydiazcheng

PR #70, opened 2024-07-10
Creating BodyWithExample Annotation

elliot-sabitov:
feat: creating BodyWithExample annotation which wraps the @Body() annotation but also allows us to pass an example body that will be used when generating the OpenAPI spec

Screenshot 2024-07-10 at 2 19 37 PM
Screenshot 2024-07-10 at 2 20 01 PM


lydiazcheng (approved):
The code looks good! 
Maybe we can consider to put mock bodies together in one place rather defined in each api code in the future.

thanaParis (approved):
approved

PR #78, opened 2024-07-11
Fix: missing energy dimension

langiole:
Added fix for energy dimension missing from second charging period.

Second charging period **before** the fix:
```
{
  "start_date_time": "2024-04-03T13:55:59.015Z",
  "dimensions": [
    {
      "type": "ENERGY_IMPORT",
      "volume": 21
    },
    {
      "type": "TIME",
      "volume": 0.0028058333333333334
    }
  ],
  "tariff_id": null
}
```

**After** the fix:
```
{
  "start_date_time": "2024-04-03T13:55:59.015Z",
  "dimensions": [
    {
      "type": "ENERGY_IMPORT",
      "volume": 21
    },
    {
      "type": "ENERGY",
      "volume": 21
    },
    {
      "type": "TIME",
      "volume": 0.0028058333333333334
    }
  ],
  "tariff_id": null
}
```


00_Base/src/mapper/session.mapper.ts
```diff
@@ -201,7 +201,7 @@ export class SessionMapper {
             });
             const previousEnergy =
               this.getEnergyImportForMeterValue(previousMeterValue);
-            if (previousEnergy) {
+            if (previousEnergy !== undefined) {
```
langiole, line 204: In this case, `previousEnergy` was coming back as 0 since the energy import of the previous charging period (the first charging period) was 0.

Therefore, this condition was returning false.

thanaParis (approved):
approved

PR #79, opened 2024-07-16
Refactor - Shared logic for Transaction map to common class and ensuring mapping during broadcast

danielyoo227:
Features:

1. Added Tariff ID on session mapping
2. Added IdToken mapping during Transaction event emit to ensure necessary information on broadcast
3. Added mapping through LocationDTO for Location ID, EVSE ID, and Connector ID
4. Refactored common logic into BaseTransactionMapper (GET Tokens, GET Locations)

Core PR:
https://github.com/citrineos/citrineos-core/pull/173


00_Base/src/mapper/session.mapper.ts
```diff
@@ -130,14 +118,40 @@ export class SessionMapper {
...
+    return connectorId ?? '';
   }
 
-  private getCurrency(location: OcpiLocation): string {
-    switch (location[OcpiLocationProps.countryCode]) {
+  private getCurrency(location: LocationDTO): string {
+    switch (location.country_code) {
       case 'US':
       default:
         return 'USD';
```
langiole, line 157: Maybe default should be null/undefined?

00_Base/src/mapper/session.mapper.ts
```diff
@@ -172,7 +189,7 @@ export class SessionMapper {
         meterValue,
         previousMeterValue,
       ),
-      tariff_id: null, // TODO: Fill in tariff_id value
+      tariff_id: tariffId, // TODO: Fill in tariff_id value
```
langiole, line 192: The TODO can be removed?

00_Base/src/mapper/BaseTransactionMapper.ts
```diff
@@ -0,0 +1,132 @@
...
+    return transactionIdToTokenMap;
+  }
+
+  protected async getTariffsForTransactions(
+    transactions: Transaction[]
+  ): Promise<Map<string, Tariff>> {
+    const transactionIdToTariffMap = new Map<string, Tariff>();
+    await Promise.all(
+      transactions.map(async (transaction) => {
+        const tariff = await this.tariffRepository.findByStationId(
```
langiole, line 76: Probably want to add a check to see if `transaction.stationId` had already been queried to avoid redundant fetched

00_Base/src/services/sessions.service.ts
```diff
@@ -33,8 +33,8 @@ export class SessionsService {
       this.transactionRepository.getTransactionsCount(dateFrom, dateTo),
     ]);
 
-    const sessions = await this.sessionMapper.mapTransactionsToSessions(
-      transactions,
+    const sessions = this.filterBasedOnCountryCodePartyId(
+      await this.sessionMapper.mapTransactionsToSessions(transactions),
```
langiole, line 37: Not something I'm asking u to do in this PR, but I would recommend we omit any references to the ocpp DBs from the service layer and instead define some sort of datasource layer.

Then, the service layer can be agnostic of the ocpp layer.

E.g.

```
  public async getSessions(
    fromCountryCode: string,
    fromPartyId: string,
    toCountryCode: string,
    toPartyId: string,
    dateFrom?: Date,
    dateTo?: Date,
    offset: number = DEFAULT_OFFSET,
    limit: number = DEFAULT_LIMIT,
  ): Promise<PaginatedSessionResponse> {
    
    const sessions = sessionsDatasource.getSessions(...);  
    const total = sessionsDatasource.getTotalSessions(...);  
    
    const response = buildOcpiPaginatedResponse(
      OcpiResponseStatusCode.GenericSuccessCode,
      total,
      limit,
      offset,
      sessions,
    );

    return response as PaginatedSessionResponse;
  }
```
The datasource layer would contain calls to the ocpp DBs and mappers as you have here.

This would allow us to define an Ocpp1.6 datasource and Ocpp2.0.1 datasource that we could easily swap in and out whenever we get to that point.

00_Base/src/services/sessions.service.ts
```diff
@@ -33,8 +33,8 @@ export class SessionsService {
       this.transactionRepository.getTransactionsCount(dateFrom, dateTo),
     ]);
 
-    const sessions = await this.sessionMapper.mapTransactionsToSessions(
-      transactions,
+    const sessions = this.filterBasedOnCountryCodePartyId(
```
langiole, line 36: Since we filter after we fetch the Sessions, wouldn't the page size potentially decrease?

00_Base/src/mapper/BaseTransactionMapper.ts
```diff
@@ -0,0 +1,141 @@
...
+  protected async getTariffsForTransactions(
+    transactions: Transaction[]
+  ): Promise<Map<string, Tariff>> {
+    const transactionIdToTariffMap = new Map<string, Tariff>();
+    const uniqueStationIds = [...new Set(transactions.map(t => t.stationId))];
+
+    const stationIdToTariffMap = new Map<string, Tariff>();
+    await Promise.all(
+      uniqueStationIds.map(async (stationId) => {
+        const tariff = await this.tariffRepository.findByStationId(stationId);
```
langiole, line 79: If it is already a part of the repo interface, we should probably do a findAll. No worries for now otherwise.

00_Base/src/mapper/BaseTransactionMapper.ts
```diff
@@ -0,0 +1,141 @@
...
+  protected async getOcpiTariffsForTransactions(
+    sessions: Session[],
+    transactionIdToTariffMap: Map<string, Tariff>,
+  ): Promise<Map<string, OcpiTariff>> {
+    const transactionIdToOcpiTariffMap = new Map<string, OcpiTariff>();
+    await Promise.all(
+      sessions.filter(session => transactionIdToTariffMap.get(session.id)).map(async (session) => {
+        const tariffKey = {
+          id: String(transactionIdToTariffMap.get(session.id)?.id),
+          // TODO: Ensure CPO Country Code, Party ID exists for the tariff in question
```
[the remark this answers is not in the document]
thanaParis: yes--in a multi-tenant system, not all tenants might require OCPI data

00_Base/src/mapper/BaseTransactionMapper.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  protected async getTariffsForTransactions(
+    transactions: Transaction[]
+  ): Promise<Map<string, Tariff>> {
+    const transactionIdToTariffMap = new Map<string, Tariff>();
+    const uniqueStationIds = [...new Set(transactions.map(t => t.stationId))];
+
+    const stationIdToTariffMap = new Map<string, Tariff>();
+    await Promise.all(
```
langiole, line 77: Why is the `Promise.all` needed here? Is the idea that each fetch is executed concurrently?

Otherwise, couldn't we have just done something like:
```
    for (const stationId of uniqueStationIds) {
      const tariff = await this.tariffRepository.findByStationId(stationId);
      if (tariff) {
        stationIdToTariffMap.set(stationId, tariff);
      }
    }
```

00_Base/src/mapper/BaseTransactionMapper.ts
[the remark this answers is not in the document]
langiole: Is there a solution here in that case?

Presumably, if we are referencing a Tariff that does not have country code + party ID, then it is a non-OCPI Tariff, and there's not very much we can do except return the Session without Tariff information which seems problematic because we can then not compute cost.

00_Base/src/mapper/BaseTransactionMapper.ts
danielyoo227: Good point, I actually did Promise.all thinking that it would run in parellel but I forgot that Javascript doesn't do that 😅 

I looked into it a bit, and it seems that Promise.all is still better to use when executing multiple promises, just due to the [way it handles the process](https://stackoverflow.com/a/54291660) (fails earlier, supports error handling better). Doesn't quite apply to this situation but learning new things every day. Adjusted the logic to do fetch ALL by the way as you pointed out, so we should be good on that regard for this PR!

elliot-sabitov: You can run the requests in parallel using Promise.all - but you cannot write it the way because currently your promises are being passed into the `.map()` function. What you need is `Promise.all(PROMISE_LIST_HERE)`... so you just need to make sure that you do a loop and put together your `const promiseList: Promise<?> = []` and then when you execute promise all it should open a separate node process for each request

elliot-sabitov: I think it is limited to the number of cores you have available in the end

00_Base/src/mapper/BaseTransactionMapper.ts
elliot-sabitov: agreed, findAll and one query would be preferred here

00_Base/src/mapper/BaseTransactionMapper.ts
danielyoo227: Oh you can! That's cool, didn't know. Is that a runtime thing? I've been reading online that you can't run parallel for node, am I just reading outdate information?

00_Base/src/services/sessions.service.ts
elliot-sabitov: I think for pagination to really work properly we need to make sure that the DB queries are written to perform operations like identifying transactions pertaining to a given partyId / countryCode combination. In other words, the pagination limit, offset, sort, etc should all be done via SQL. Ideally, if we were to be able to query DB for all new transactions where we join the location, charging station, ocpi location, etc, tables and in one query can fetch latest set of transactions via the query then I think we will be closer to having pagination working better. Otherwise, what tends to happen is you end up getting a sub set of data and cant properly have pagination, or are forced to fetch all data and have some sort of local cache layer.

Since we are using `sequelize-typescript` you would have to follow their syntax similar to what we have here:
```
await this.transactionEventRepository.readAllTransactionsByQuery({
          where: {
            stationId: stationId,
            isActive: true,
          },
          include: [
            {
              model: Evse,
              where: {
                id: evseId,
              },
            },
          ],
        })
```
where you can see the `Evse` is joined and we query based on evse id. @danielyoo227 what are your thoughts on this?

elliot-sabitov: it is worth noting with the `include: [...]` array, you can recursively include nested models and in any level down the chain you can add a `where`

danielyoo227: @elliot-sabitov Yep, 100%. Similar change coming up to ensure that we filter everything at DB level for country code and party ID, along with the pagination params. It'll join the OCPI Token and OCPI Locations data to check on conutry code and party ID before returning

00_Base/src/mapper/BaseTransactionMapper.ts
elliot-sabitov: Promise.all should trigger all requests in a non blocking fashion and wait for all requests to complete before returning, so technically they will be running in parallel. I think you can also start a new node process if you really wanted it to run on another thread on your processor though which can also be done by doing like `exec('npm run something-else`)` but I think Promise.all should suffice in triggering the requests in parallel

00_Base/src/services/cdrs.service.ts
```diff
@@ -1,15 +1,16 @@
...
+import { ICdrsDatasource } from '../datasources/ICdrsDatasource';
+import { CDR_DATASOURCE_SERVICE_TOKEN } from '../datasources/CdrsDatasource';
 
 @Service()
 export class CdrsService {
   constructor(
-    private readonly transactionRepository: SequelizeTransactionEventRepository,
-    private readonly cdrMapper: CdrMapper,
-  ) {}
+    @Inject(CDR_DATASOURCE_SERVICE_TOKEN)
```
langiole, line 10: Does doing this work: `@Inject(() => CdrsDatasource)`? Seems cleaner.

00_Base/src/datasources/CdrsDatasource.ts
```diff
@@ -0,0 +1,29 @@
...
+import { ITransactionDatasource } from './ITransactionDatasource';
+import { TRANSACTION_DATASOURCE_SERVICE_TOKEN } from './TransactionDatasource';
+import { ICdrsDatasource } from './ICdrsDatasource';
+import { CdrMapper } from '../mapper/cdr.mapper';
+import { Cdr } from '../model/Cdr';
+
+export const CDR_DATASOURCE_SERVICE_TOKEN = new Token('CDR_DATASOURCE_SERVICE_TOKEN');
+
+@Service(CDR_DATASOURCE_SERVICE_TOKEN)
+export class CdrsDatasource implements ICdrsDatasource {
```
langiole, line 12: Can leave as is for now, but in future we might want to consider renaming to something like `CdrsDatasourceOcpp2_0_1`.

Thinking ahead, for a network that is both 1.6 and 2.0.1, I'm just wondering how the get pagination logic would work since it would need to fetch and merge both 1.6 and 2.0.1 data. Seems like it will be tricky.

Not something we need to worry about rn though.

00_Base/src/datasources/CdrsDatasource.ts
```diff
@@ -0,0 +1,29 @@
...
+  constructor(
+    @Inject(TRANSACTION_DATASOURCE_SERVICE_TOKEN)
+    private readonly transactionDatasource: ITransactionDatasource,
+    private readonly cdrMapper: CdrMapper
+  ) {
+  }
+
+  async getCdrs(cpoCountryCode: string, cpoPartyId: string, mspCountryCode: string, mspPartyId: string, dateFrom?: Date, dateTo?: Date, offset?: number, limit?: number): Promise<PaginatedResult<Cdr>> {
```
langiole, line 20: Your call, but I kind of like the return type that I've seen @yuvenus use where we return something like `[Cdr[], int]`.

Then, on the other side, it would look like: `const [cdrs, total] = datasource.getCdrs(...)`.

But what you have rn is nice cause of how explicit it is so idk.

00_Base/src/datasources/ITransactionDatasource.ts
langiole, line 1: Not sure we want this datasource, as the point of the datasource was to abstract away the fetching of OCPI models of which `Transaction` is not OCPI.

00_Base/src/services/cdrs.service.ts
```diff
@@ -21,30 +22,33 @@ export class CdrsService {
...
-    response.total = total;
+    response.data = result.data;
+    response.total = result.total;
     response.offset = offset;
     response.limit = limit;
 
     return response;
   }
+
+  private filterBasedOnCountryCodePartyId(
```
langiole, line 36: No usages?

langiole (approved):
Looks good! Left some minor feedback

00_Base/src/mapper/cdr.mapper.ts
```diff
@@ -1,62 +1,53 @@
...
-          transactionIdToTariffMap,
-          toCountryCode,
-          toPartyId,
+          transactionIdToTariffMap
         );
       return await this.mapSessionsToCDRs(
         sessions,
+        transactionIdToLocationMap,
         transactionIdToTariffMap,
         transactionIdToOcpiTariffMap,
```
langiole, line 52: For simplicity sake, I would love if this could be as simple as "Build CDRs given OCPI Locations and OCPI Tariffs".

But instead, we pass in both Core Tariffs and OCPI Tariffs.

danielyoo227: Makes sense, will track with ticket

00_Base/src/datasources/ITransactionDatasource.ts
danielyoo227: We do need something here so that we can fetch transactions by country code and party ID 🤔 Should we name it something else and put it as a service?

langiole: I might be misunderstanding, but I don't think we should be storing OCPI related fields (country code and party ID) in Transactions (Core models in general).

We probably should be storing an OCPI Session model that holds country code + party ID + transaction ID, and should do a fetch on this first in order to fetch Transactions.

Maybe we can discuss further in daily.

danielyoo227:
> @danielyoo227 @elliot-sabitov @yuvenus @langiole This PR introduces 'datasource' classes. I see here we have Sessions and Cdrs datasources. I don't feel good about making this refactor only for some modules. What is the LoE for incorporating this datasource pattern across the modules?

@thanaParis I see 7 other OCPI modules that would need to be refactored to support this. Assuming that they follow similar fetch patterns to Sessions and CDRs, I can estimate 1-2 hours for each module for refactor / testing. So I'd assume 1-2 days for the effort to incorporate the pattern across the modules. I can take this on and add it to this PR if we'd like to do this now.

00_Base/src/mapper/BaseTransactionMapper.ts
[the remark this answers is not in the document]
thanaParis: This is a good point. I think the validation should happen elsewhere... it should be impossible for a connector to lack a tariffId, and it should be impossible for a tariffId on a connector to not be an ocpi tariff (or for an ocpi tariff to be 'downgraded' to a non-ocpi tariff). If these things are true, this logic should never break.

PR #81, opened 2024-07-16
Fix Sessions Broadcaster Push PUT vs PATCH

elliot-sabitov:
fix: adjusting session broadcaster to send PUT vs PATCH push to clients depending on whether the transactionRepository fires a 'created' or 'updated' event to ensure that it does not appear as a duplicate PUT
Screenshot 2024-07-16 at 12 40 34 PM
Screenshot 2024-07-16 at 12 40 28 PM


00_Base/src/broadcaster/session.broadcaster.ts
```diff
@@ -22,34 +29,57 @@ export class SessionBroadcaster extends BaseBroadcaster {
...
+        );
+        this.broadcast(transactions, PutOrPatch.PATCH).then();
+      }
+    });
   }
 
-  private async broadcast(transactions: Transaction[]) {
+  private async broadcast(
+    transactions: Transaction[],
+    putOrPatch: PutOrPatch = PutOrPatch.PUT,
```
langiole, line 54: why not just have boolean `patch`?

E.g.
```
  private async broadcast(
    transactions: Transaction[],
    patch: boolean = false,
  )
```

elliot-sabitov:
Removed WIP after improving implementation such that for PATCH request we no subscribe to creation of transaction events instead of transactions, which enables us to get the meterValues/charging periods, and kwh, totalCost and lastUpdated from the associated transaction and only patch those four values. This resolves the issue where all meter values were previously included, now, only the meter values for that one transaction event will be included

Screenshot 2024-07-17 at 2 44 31 PM
Screenshot 2024-07-17 at 2 44 25 PM

00_Base/src/broadcaster/session.broadcaster.ts
```diff
@@ -22,34 +31,102 @@ export class SessionBroadcaster extends BaseBroadcaster {
...
+      if (transactions && transactions.length > 0) {
+        this.logger.info(
+          'Attempting to broadcast created transactions',
+          transactions.map((t) => t.id),
+        );
+        this.broadcast(transactions).then();
+      }
+    });
+    this.transactionRepository.on(
+      'created',
```
yuvenus, line 44: should this be 'updated' instead?

00_Base/src/broadcaster/session.broadcaster.ts
elliot-sabitov: adjusted

00_Base/src/broadcaster/session.broadcaster.ts
elliot-sabitov: interesting that you point that out because that was my first thought as well - but it seems that the way we have it is that there is one `Transaction` which has multiple `TransactionEvents`, and we use the `TransactionEventRepository` to access those transaction events. The partially confusing part is that the `TransactionEventRepository` has `transaction: TransactionRepository` on it, which is why here we have to subscribe to the creation of the TransactionEvents that come in because they don't really seem to get updated

langiole: @elliot-sabitov 

You're saying that this is subscribing to creation of TransactionEvents?

Maybe we should consolidate these into methods: `subscribeToTransactionCreated()` and `subscribeToTransactionEventCreated()`?

00_Base/src/broadcaster/session.broadcaster.ts
```diff
@@ -22,34 +25,99 @@ export class SessionBroadcaster extends BaseBroadcaster {
...
       cpoCountryCode,
       cpoPartyId,
       ModuleId.Sessions,
       params,
-      this.sessionsClientApi.putSession.bind(this.sessionsClientApi),
+      clientApiRequest.bind(this.sessionsClientApi),
     );
   }
+
+  private getSessionForPatch(session: Session): Partial<Session> {
```
langiole, line 115: Minor, but maybe `buildSessionPatch` is better name?

00_Base/src/broadcaster/session.broadcaster.ts
```diff
@@ -22,34 +25,99 @@ export class SessionBroadcaster extends BaseBroadcaster {
...
+    return {
+      kwh: session.kwh,
+      charging_periods: session.charging_periods,
```
langiole, line 118: So we'd be sending the full charging periods? If so, seems to defeat the purpose of the patch.

00_Base/src/broadcaster/session.broadcaster.ts
```diff
@@ -22,34 +25,99 @@ export class SessionBroadcaster extends BaseBroadcaster {
...
+    }
     const cpoCountryCode = session.country_code;
     const cpoPartyId = session.party_id;
     await this.sessionsClientApi.broadcastToClients(
       cpoCountryCode,
       cpoPartyId,
       ModuleId.Sessions,
       params,
-      this.sessionsClientApi.putSession.bind(this.sessionsClientApi),
+      clientApiRequest.bind(this.sessionsClientApi),
```
langiole, line 111: Could this arg have not have been some sort of `isPatch` boolean?

Passing in this `bind` seems odd.

00_Base/src/broadcaster/session.broadcaster.ts
langiole, line 1: I was a little bit confused at first by the varying broadcast related method declarations since we have `broadcast`, `broadcastPatch`, and `sendSessionToClients`.

Lmk what you think of this refactor:

```
@Service()
export class SessionBroadcaster extends BaseBroadcaster {
  constructor(
    readonly logger: Logger<ILogObj>,
    readonly transactionRepository: SequelizeTransactionEventRepository,
    readonly sessionMapper: SessionMapper,
    readonly sessionsClientApi: SessionsClientApi,
    readonly credentialsService: CredentialsService,
  ) {
    super();
    this.transactionRepository.transaction.on('created', (transactions) => {
      if (transactions && transactions.length > 0) {
        this.logger.info(
          'Attempting to broadcast created transactions',
          transactions.map((t) => t.id),
        );
        this.broadcastTransactions(transactions).then();
      }
    });
    this.transactionRepository.on(
      'created',
      async (transactionsEvents: TransactionEvent[]) => {
        if (transactionsEvents && transactionsEvents.length > 0) {
          this.logger.info(
              'Attempting to broadcast updated transactions',
              transactionsEvents.map((t) => t.id),
          );
          const transactions = await this.buildTransactions(transactionsEvents);
          if (transactions.length === 0) {
            return Promise.resolve(); // skipping because no transaction not yet created for transaction event
          } else {
            return this.broadcastTransactions(transactions, true);
          }
        }
      },
    );
  }

  private async buildTransactions(transactionsEvents: TransactionEvent[]) {
    const transactions: Transaction[] = [];
    for (const transactionsEvent of transactionsEvents) {
      if (
          transactionsEvent.meterValue &&
          transactionsEvent.meterValue.length > 0 &&
          transactionsEvent.triggerReason !== TriggerReasonEnumType.RemoteStart // skip remote start because it will be handled by PUT
      ) {
        // skip if there are no meter values in transaction event
        const transaction: Transaction = (await transactionsEvent.$get(
            'transaction',
        )) as Transaction; // todo use Props
        if (transaction) {
          transaction.setDataValue('transactionEvents', [transactionsEvent]); // todo use Props
          transaction.setDataValue('meterValues', transactionsEvent.meterValue); // todo use Props
          transaction.transactionEvents = [transactionsEvent];
          transaction.meterValues = transactionsEvent.meterValue;
          transactions.push(transaction);
        }
      }
    }
    return transactions;
  }

  private async broadcastTransactions(transactions: Transaction[], isPatch = false) {
    // todo do we know if we can do put vs patch here, is there a way to have a delta?
    const sessions: Session[] =
      await this.sessionMapper.mapTransactionsToSessions(transactions);

    for (const session of sessions) {
      await this.broadcastSession(session, isPatch);
    }
  }

  private async broadcastSession(
    session: Session,
    isPatch = false,
  ): Promise<void> {
    let params: any = PutSessionParams.build(session.id, session);
    let clientApiRequest: any = this.sessionsClientApi.putSession;
    if (isPatch) {
      params = PatchSessionParams.build(
        session.id,
        this.getSessionForPatch(session),
      );
      clientApiRequest = this.sessionsClientApi.patchSession;
    }
    const cpoCountryCode = session.country_code;
    const cpoPartyId = session.party_id;
    await this.sessionsClientApi.broadcastToClients(
      cpoCountryCode,
      cpoPartyId,
      ModuleId.Sessions,
      params,
      clientApiRequest.bind(this.sessionsClientApi),
    );
  }

  private getSessionForPatch(session: Session): Partial<Session> {
    return {
      kwh: session.kwh,
      charging_periods: session.charging_periods,
      total_cost: session.total_cost,
      last_updated: session.last_updated,
    };
  }
}
```

langiole (approved):

00_Base/src/broadcaster/session.broadcaster.ts
elliot-sabitov: adjusted

00_Base/src/broadcaster/session.broadcaster.ts
elliot-sabitov: adjusted

00_Base/src/broadcaster/session.broadcaster.ts
elliot-sabitov: it is my understanding that we will be sending only the charging periods that come from the meter values that are in the transaction event so it should only be the latest charging periods right?

00_Base/src/broadcaster/session.broadcaster.ts
elliot-sabitov: the `broadcastToClients` is a generic method in `BaseClientApi` which is intended to take in a request that needs to be performed... we probably could have overrode `broadcastToClients` in `SessionsClientApi` potentially to handle the boolean, but I think this may be sufficient, but let me know your thoughts

00_Base/src/broadcaster/session.broadcaster.ts
elliot-sabitov: adjusted - let me know what you think!

langiole (approved):

danielyoo227 (approved):

PR #82, opened 2024-07-17
Feature: Registration - Generate Credential Token A

lydiazcheng:
## Changes
1. Add an admin endpoint to generate credentials token A (Reference: OCPI 2.2.1, 7.1.1. Registration)
2. Small fixes

## Tests
**1. Call generate token A endpoint**
- We expect token A is in the response Screenshot 2024-07-16 at 8 23 06 PM
- We expect new CpoTenant are created Screenshot 2024-07-16 at 8 23 20 PM
- We ServerCredentialRole are created Screenshot 2024-07-16 at 8 23 42 PM
- We expect ClientInformation with server token(token A) are created but not registered Screenshot 2024-07-16 at 8 23 59 PM
- We expect ServerVersion and its endpoints are created Screenshot 2024-07-16 at 8 24 23 PM Screenshot 2024-07-16 at 8 24 37 PM

**2. MSP sends get versions request to OCPI using token A and url from previous response** Screenshot 2024-07-16 at 8 26 23 PM

**3. MSP sends get version details request using token A and the url from previous response** Screenshot 2024-07-16 at 8 27 57 PM

**4. MSP posts credentials token B (5aa45990-7600-44b3-9944-d3ea855dd81f) using token A and credentials url from the previous response**
- We expect OCPI send get version and version details request to MSP using the token B Screenshot 2024-07-16 at 8 34 31 PM
- We expect a new token C is generated and sent in response Screenshot 2024-07-16 at 8 35 14 PM
- We expect token B and token C are registered and stored in the clientInformation which we just created in step 1 Screenshot 2024-07-16 at 8 36 13 PM
- We expect client credential role are stored Screenshot 2024-07-16 at 8 36 56 PM
- We expect client version details is stored Screenshot 2024-07-16 at 8 37 29 PM 
- We expect client version endpoints are stored based on the response in get version details from MSP 
Screenshot 2024-07-16 at 8 40 12 PM Screenshot 2024-07-16 at 8 40 55 PM


00_Base/src/services/credentials.service.ts
```diff
@@ -474,6 +477,119 @@ export class CredentialsService {
...
+    const clientInfo = await ClientInformation.create(
+      {
+        cpoTenantId: cpoTenant.id,
+        registered: false,
+        serverToken: credentialsTokenA,
+        serverVersionDetails: [
+          {
+            version: versionNumber,
+            url: credentialsRequest.url,
+            endpoints: storedVersion.endpoints.map((endpoint) => ({
```
lydiazcheng, line 564: Hi @elliot-sabitov, in the discussion yesterday we decided not to add endpoints here. But then in the test I found after finishing the whole registration process, these endpoints were not get added. The later requests are focus on client versions. So I still added endpoints here. Please feel free let me know if I get it wrong. Thank you!

00_Base/src/model/ServerCredentialsRole.ts
```diff
@@ -67,3 +82,18 @@ export class ServerCredentialsRole extends Model implements ICredentialsRole {
     return serverCredentialsRole;
   }
 }
+
+export const toCredentialsRoleDTOFromServerCredentialsRole = (
```
langiole, line 86: Why not make this method a part of `ServerCredentialsRole` class?

Then you can make call: `serverCredentialsRole.toCredentialsRoleDto()`.

00_Base/src/services/credentials.service.ts
```diff
@@ -546,15 +662,9 @@ export class CredentialsService {
...
-    const existingVersionDetails = clientInformation.clientVersionDetails.find(
-      (v: ClientVersion) => v.version === versionNumber,
-    );
-    if (!existingVersionDetails) {
-      throw new NotFoundError('Version details not found');
-    }
     const clientVersion = await this.getVersionDetails(
       versionNumber,
-      existingVersionDetails.url,
+      credentials.url,
```
thanaParis, line 667: this method is getting VersionDetails--the credentials.url gets Versions
the VersionDetail url is in the Version object, which is why the previous lookup was happening

00_Base/src/services/credentials.service.ts
```diff
@@ -474,6 +477,119 @@ export class CredentialsService {
...
+            role.country_code,
+            role.party_id,
+          );
+        if (storedRole) {
+          throw new InternalServerError(
+            `ServerCredentialsRole with country_code ${role.country_code} and party_id ${role.party_id} already exists.`,
+          );
+        }
+      } catch (e) {
+        if (e instanceof NotFoundError) {
```
langiole, line 499: Personal preference but I wish we did not have our data layer throw exception when something is not found.

I think returning `undefined` would result in cleaner code.

00_Base/src/services/credentials.service.ts
thanaParis, line 667: this method is getting VersionDetails--the credentials.url gets Versions
the VersionDetail url is in the Version object, which is why the previous lookup was happening

00_Base/src/services/credentials.service.ts
```diff
@@ -474,6 +477,119 @@ export class CredentialsService {
...
+
+    for (const role of serverCredentialsRoleDTOs) {
+      try {
+        const storedRole =
+          await this.serverCredentialsRoleRepository.getServerCredentialsRoleByCountryCodeAndPartyId(
+            role.country_code,
+            role.party_id,
+          );
+        if (storedRole) {
+          throw new InternalServerError(
```
langiole, line 494: Maybe more of `BadRequestError` instead?

thanaParis, line 494: why does this throw an error? the server's cpo role would be reused across many credentials, no?

00_Base/src/services/credentials.service.ts
```diff
@@ -474,6 +477,119 @@ export class CredentialsService {
     return updatedClientInformation;
   }
 
+  async generateCredentialsTokenA(
+    credentialsRequest: CredentialsRequestDTO,
+    versionNumber: VersionNumber,
+  ): Promise<CredentialsDTO> {
+    const serverCredentialsRoleDTOs = credentialsRequest.roles;
+
+    for (const role of serverCredentialsRoleDTOs) {
```
langiole, line 486: Maybe consider wrapping this for loop in a method `validateServerRolesForRegistration`?

I would also think that the logic for `registerCredentialsTokenA` should also have this same validation logic, no?

thanaParis (changes requested):

00_Base/src/services/credentials.service.ts
```diff
@@ -474,6 +477,119 @@ export class CredentialsService {
     return updatedClientInformation;
   }
 
+  async generateCredentialsTokenA(
```
langiole, line 480: I might be missing something but shouldn't the logic for `generateCredentialsTokenA` and `registerCredentialsTokenA` be very similar?

The only difference should be that `generateCredentialsTokenA` generates a new token while `registerCredentialsTokenA` uses the token provided in the request.

If so, then I would expect the logic in `generateCredentialsTokenA` to be as simple as:
1. Generate token
2. Make call to `registerCredentialsTokenA`

00_Base/src/services/credentials.service.ts
lydiazcheng: Hi Thana, thanks for the review! 

This `getVersionDetails` method actually get versions at first and then get version details. The input url is used to get version so I changed it to use this `credentials.url `. 

I also removed the previous lookup for `existingVersionDetails` because when we initially receives the postCredential call from MSP, there has not been client side version details. We need to use the credential url from the request body to get client side versions and then details (we do it inside this `getVersionDetails` method). The previous lookup and throw exceptions make the process fail.

00_Base/src/services/credentials.service.ts
lydiazcheng: I changed the logic to update if the role exists but still validate if roles from the request body has been stored then they should belongs to the same CpoTenant. Please feel free to let me know if we should do it in another way.

00_Base/src/services/credentials.service.ts
lydiazcheng: Hi Lee, thanks for the review! 

+ 1. Currently this `getServerCredentialsRoleByCountryCodeAndPartyId` is used in multiple places. We might do an improvement later.

00_Base/src/services/credentials.service.ts
lydiazcheng: `registerCredentialsTokenA` doesn't has the save validation in our current code and we assume we should always create new objects since the register process happens at the very beginning. Also based on the OCPI protocol, when registering, receiver (MSP) only send us version url and token A. So we should not need roles. We might refactor the request body of the register API.

00_Base/src/services/credentials.service.ts
lydiazcheng: The registration process is symmetric (refer OCPI 7.1.1. Registration). OCPI can be the sender or the receiver.

When OCPI is the sender and MSP is the receiver, we call `registerCredentialsTokenA`. The token A is from MSP and is a temp token which is only valid during the registration process(MSP sends us token A by email or any other safe way). In the `registerCredentialsTokenA`, we use this token A to get MSP side version, version details and send them our token B. In the response, MSP send us the client roles as well as formal token C which replaces the temp token A. The registration process finishes.

When OCPI is the receiver and MSP is the sender,  we call `generateCredentialsTokenA` to generate our side temp token A and prepare server side info, e.g., server roles, versions and details. Then we send this token A to MSP also by emails or other ways. This time MSP execute the registration like we do in the `registerCredentialsTokenA`. 

So we can not simply do generate token and call `registerCredentialsTokenA` together. They serves for different cases as sender and receiver.

00_Base/src/model/DTO/CredentialsRequestDTO.ts
```diff
@@ -0,0 +1,35 @@
...
+  IsArray,
+  IsNotEmpty,
+  IsString,
+  IsUrl,
+  ValidateNested,
+} from 'class-validator';
+import { Type } from 'class-transformer';
+import { CredentialsRoleDTO } from './CredentialsRoleDTO';
+
+export class CredentialsRequestDTO {
```
elliot-sabitov, line 12: may be helpful to have Admin in the name to separate from the other credential endpoints

00_Base/src/services/credentials.service.ts
```diff
@@ -474,6 +478,109 @@ export class CredentialsService {
...
+    credentialsRequest: CredentialsRequestDTO,
+    versionNumber: VersionNumber,
+  ): Promise<CredentialsDTO> {
+    const serverCredentialsRoleDTOs = credentialsRequest.roles;
+
+    // Make sure we stored the necessary version and version endpoints
+    // so that MSP can retrieve them later in the registration process
+    const storedVersion = await this.versionRepository.readOnlyOneByQuery(
```
elliot-sabitov, line 489: can we move this to a reusable validator, or better yet maybe we can do this in middleware? It would be nice to have a single method that can validate that the version number is supported for a given ModuleId...

00_Base/src/services/credentials.service.ts
```diff
@@ -474,6 +478,109 @@ export class CredentialsService {
...
+            include: [Endpoint],
+          },
+        ],
+      },
+    );
+
+    const storedServerCredentialsRoleDTOs = storedServerCredentialsRoles.map(
+      (storedRole) => storedRole.toCredentialsRoleDTO(),
+    );
+    return CredentialsDTO.build(
```
elliot-sabitov, line 577: I see that in this method you have validated that we support the version, got existing tenant or created one if there wasnt one existing, then added server credentials with provided CredentialRole information and created the new ClientInformation with the newly generated `serverToken` which can be used by the sender for auth, and finally you return the CredentialsDTO in the response. All looks good - good job 👍

elliot-sabitov: it may be helpful to separate some of these chunks of code into smaller functions

00_Base/src/services/credentials.service.ts
elliot-sabitov: yep, the one helper `this.getVersionDetails(` is designed to return VersionDetails which is created after performing both of the get versions and the get version details requests

elliot-sabitov: oh wait yeah @lydiazcheng, Thana is right, as I just now pointed out, you need to pass in the versions endpoint url here and not the credentials endpoint because that function will perform both GET versions requests

00_Base/src/services/credentials.service.ts
```diff
@@ -656,4 +757,41 @@ export class CredentialsService {
...
     return await clientInformation.save();
   }
+
+  private async getCpoTenantIdForServerRolesForRegistration(
+    serverCredentialsRoleDTOs: CredentialsRoleDTO[],
+  ): Promise<number | undefined> {
+    let cpoTenantId: number | undefined;
+    for (const role of serverCredentialsRoleDTOs) {
+      try {
+        if (role.role !== Role.CPO) {
```
elliot-sabitov, line 767: I suggest trying to perform all validation in the beginning prior to performing any queries or anything like that. I understand that intuitively we catch errors as we write the code, but in terms of operation execution, your code will benefit from being structured in a way that all validation that is needed in the endpoint handler happens first, prior to execution, which I can understand in some cases may be unavoidable, but I can see that you actually do handle all the error with payload in version right in the beginning. I would suggest creating a `validate` method which than triggers a few more `validateVersion`, `validateRole`, etc. smaller methods. Each method name and definition helps the reader quicker understand what is going on and having smaller methods also helps with testability.

I have an issue with this as well 🤦

00_Base/src/model/ServerCredentialsRole.ts
elliot-sabitov: agreed - but also please make it `static`.

Please note that there may be issues if your class extends Model and you are trying to define a `static build(..)` method because of conflicts. In the majority of other cases where `static` can be used, we should. In some cases, static context is limited for... reasons... and this approach can be used as an alternative. In some cases, you may want to move a `const fun() => {}` into a util or something so that it can be imported and reused without having the import the rest of the class, especially if that class extends Model or has some other imports that you are trying to avoid.

I think this may also have been something that we had previously done for seeders because they didnt have the same Models imported so we would get a sequelize error - but now we saw that we can resolve that by making sure that `DefaultSequelizeInstance` or `OcpiSequelizeInstance` is initialized, whichever applicable.

00_Base/src/services/credentials.service.ts
lydiazcheng: Hi @elliot-sabitov, the passed in url is only used to get versions. The `baseUrl` is then replaced by the real version url later after we receive the response of get versions (this [line](https://github.com/citrineos/citrineos-ocpi/blob/017d7d8093d44aa4b74f8461bac75bc0ef056224/00_Base/src/services/credentials.service.ts#L585)). 

And we are not able to pass in any version endpoint url into this `getVersionDetails` because we don't have any client side `existingVersionDetails` at the moment when calling this `getVersionDetails`. As described in OCPI Registration process, when we play as a receiver, the MSP send us postCredential request, then we start to send get version and get the version details request to get client side version details.

elliot-sabitov: ah I see right okay, makes sense

00_Base/src/services/credentials.service.ts
lydiazcheng: Hi @elliot-sabitov, thanks for the feedback! I updated the code in this commit: 897073ec23ae680316e4f6b7ca76fdc50eee4544. Please feel free to let me know if there is any further comments.

Regarding this version validation, I move it partially into a validator script since I still need all version endpoints in the following step. We might not need to do it in a middleware for every module additionally. Because version module has no delete actions defined in OCPI protocol. As long as cpo and msp exchange version endpoints that each other has during the registration, it should be safe.

00_Base/src/model/ServerCredentialsRole.ts
```diff
@@ -77,4 +78,22 @@ export class ServerCredentialsRole extends Model implements ICredentialsRole {
     serverCredentialsRole.business_details = businessDetails;
     return serverCredentialsRole;
   }
+
+  static toCredentialsRoleDTO(
+    serverCredentialsRole: ServerCredentialsRole,
+  ): CredentialsRoleDTO {
+    const credentialsRoleDTO = new CredentialsRoleDTO();
+    credentialsRoleDTO.role =
+      serverCredentialsRole[ServerCredentialsRoleProps.role];
```
langiole, line 87: I would say that if we don't need to use the "Props" then maybe we shouldn't for the sake of readability.

elliot-sabitov: the issue is that we need props because the lazy loading with `Model.$get('fieldName')` takes in that field name string and if I keep it as a string, then in the future if the model is to change, it will introduce bugs, so there needs to be some sort of constants for sequelize models =/ but using Props in general helps in other situations as well as it helps ensure that the property name is centralized...

lydiazcheng: If we use the way of `model.fieldName` here, it is easy to read. And when the fieldName is changed, there will be a build error so that we still have a chance to notice it. From this perspective, it is safer than `Model.$get('fieldName')`. But using props will save more effort, i.e, we just need to change one place when update fieldName. Personally, I would prefer to using props.

elliot-sabitov: Using `mode.fieldName` won't lazy load and query the DB to get the association thought right? I think you do need `.$get(..)` at least in some places where it makes sense... Also, wouldn't `.$get(..)` only error out during runtime? I was thinking that this could introduce a hard to find bug that will only happen when this code runs..

thanaParis (approved):
Approved

PR #86, opened 2024-07-17
HTTP Exception Handler for Admin Endpoints

yuvenus:
This PR creates a separate exception handler called `HttpExceptionHandler` that returns a generic response in the event of an error, rather than the bespoke OCPI response bodies required of the main OCPI endpoints. This error handler is incorporated into the new decorator `AsAdminEndpoint` that will be used to support any admin endpoint's error handling.

The original `GlobalExceptionHandler` was renamed to `OcpiExceptionHandler` as a result of the change above.

### DIFF
--- a/00_Base/src/util/decorators/as.ocpi.functional.endpoint.ts
+++ b/00_Base/src/util/decorators/as.ocpi.functional.endpoint.ts
@@ -10,6 +10,7 @@ import { OcpiHttpHeader } from '../ocpi.http.header';
 import { OcpiHeaderMiddleware } from '../middleware/ocpi.header.middleware';
 import { UniqueMessageIdsMiddleware } from '../middleware/unique.message.ids.middleware';
 import { HttpHeader } from '@citrineos/base';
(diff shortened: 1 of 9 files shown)


00_Base/src/util/decorators/as.ocpi.functional.endpoint.ts
```diff
@@ -36,5 +37,6 @@ export const AsOcpiFunctionalEndpoint = function () {
     UseBefore(AuthMiddleware)(object, methodName);
     UseBefore(OcpiHeaderMiddleware)(object, methodName);
     UseBefore(UniqueMessageIdsMiddleware)(object, methodName);
+    UseBefore(OcpiExceptionHandler)(object, methodName);
```
elliot-sabitov, line 40: haha, I like how you put it on the other decorators - good call 👍

elliot-sabitov (approved):

thanaParis (approved):
Approved!

PR #88, opened 2024-07-18
Admin Create/Update Locations Endpoint

yuvenus:
**Matching Core PR:** https://github.com/citrineos/citrineos-core/pull/172

This PR adds the ability to create/update locations via an "admin" endpoint found in Locations module.

### Changes:

- Added new service `admin.ts` in Locations module to handle "admin"-related functionality, such as creating/updating a location 
- Added new endpoint `PUT /admin` that handles both create and update
  - Endpoint requires `broadcast` query param to know if the location should be broadcasted to MSPs
  - Endpoint will respond with the created/updated location object
- Added new DTO called `AdminLocationDTO` to be a dedicated request body for admin endpoint, since there is information not derivable from the "regular" location DTO
  - Includes `AdminEVSEDTO` and `AdminConnectorDTO`
  - TODO in another PR to allow more OCPI information into this DTO for creation
- Updated MSP broadcast to always do a PUT for now, regardless of create or update
- Moved variable attribute map creation into its own file called `VariableAttributesUtil` so that locations and locations admin services can benefit from the same logic
- Moved OCPI-info map creation into its own file called `OcpiLocationsUtil` so that locations and locations admin services can benefit from the same logic
- Replaced error throwing with simple logging in mapper to not "punish" other locations for one incomplete location
- Removed `IOcpiLocationMapper` interface because it was not used for anything
- Removed 'ocpi.evse.repository.ts` since it's now replaced by the "real" `OcpiEvseRepository`
- Added `@types/geojson` dependency to make coordinate processing a little neater

### How the admin locationsendpoint works:

- The endpoint is at `PUT /ocpi/2.2.1/locations/admin`
- If the request body does not include an id:
  - Validate that the request body has the minimal amount of information to create the location/evse/connector
    - If any layer is missing required information, return an error specifying which information is required 
  - Make both the OCPI location and Citrine location, without checking if one already exists
- If the request body includes an id:
  - Validate that the request body has the minimal amount of information to create the evse/connector (location is good at this point)
      - If any layer is missing required information, return an error specifying which information is required
  - Create/update the Citrine Location
  - Create/update the Charging Stations attached to the Citrine Location
  - Create/update the OCPI Location based on the Citrine Locationid
  - Create/update the OCPI EVSE and OCPI Connector information
- Retrieve all relevant variable attributes
- Retrieve all relevant OCPI information
- Map to LocationDTO
- If `broadcast`, send Location to MSPs
- Return mapped LocationDTO


00_Base/src/model/DTO/admin/AdminLocationDTO.ts
```diff
@@ -0,0 +1,152 @@
...
+  connectors?: AdminConnectorDTO[];
+}
+
+export class AdminConnectorDTO {
+  @IsNumber()
+  @IsNotEmpty()
+  id!: number;
+}
+
+export const isLocationInvalid = (
```
elliot-sabitov, line 108: can we make this a static method on the DTO class?

00_Base/src/repository/OcpiLocationRepository.ts
```diff
@@ -77,6 +77,54 @@ export class OcpiLocationRepository extends SequelizeRepository<OcpiLocation> {
...
+            [OcpiLocationProps.countryCode]:
+              location[OcpiLocationProps.countryCode],
+            [OcpiLocationProps.publish]: location[OcpiLocationProps.publish],
+            [OcpiLocationProps.lastUpdated]:
+              location[OcpiLocationProps.lastUpdated],
+            [OcpiLocationProps.timeZone]: location[OcpiLocationProps.timeZone],
+          },
+        });
+      if (!ocpiLocationCreated) {
+        const values: Partial<OcpiLocation> = {};
```
elliot-sabitov, line 103: please initialize a new OcpiLocation

elliot-sabitov (approved):

00_Base/src/model/DTO/admin/AdminLocationDTO.ts
yuvenus: done!

03_Modules/Locations/src/module/admin.ts
langiole, line 1: We should probably be consistent with where we stick our admin services.

I have been putting them inline with the other services in base.

I don't mind putting them within each module, but maybe we should wait to do so until we decide how to move the current services into their respective modules?

yuvenus: i moved the admin functionality out of the locations service because the file was getting far too big 😰 additionally i put the admin service solely in Locations module because i'm not aware of any other uses of the admin locations functionality, but if we reach the consensus that all services belong in 00_Base, regardless of usability between modules, i can definitely move the admin service in there 👍

00_Base/src/util/OcpiLocationsUtil.ts
```diff
@@ -0,0 +1,65 @@
...
+    private ocpiConnectorRepository: OcpiConnectorRepository,
+  ) {}
+
+  public async createOcpiEvsesInfoMap(
+    chargingStationAttributesMap: Record<
+      string,
+      ChargingStationVariableAttributes
+    >,
+  ): Promise<Record<string, OcpiEvse>> {
+    const ocpiEvseMap: Record<string, OcpiEvse> = {};
```
langiole, line 23: Could we have used the `Map` type instead of `Record`?

03_Modules/Locations/src/module/admin.ts
```diff
@@ -0,0 +1,172 @@
...
+    adminLocationDto: AdminLocationDTO,
+    broadcast: boolean,
+  ): Promise<LocationDTO> {
+    this.logger.debug(
+      adminLocationDto.id
+        ? `Updating Location ${adminLocationDto.id}`
+        : `Creating Location ${adminLocationDto.name}`,
+    );
+
+    const [invalidLocation, message] = AdminLocationDTO.IS_LOCATION_INVALID(adminLocationDto);
```
langiole, line 51: Should `IS_LOCATION_INVALID` be in standard camel case since it is more of a method?

Also, why not make this a non static method? That way we can write:
`adminLocationDto.isValid()`?

03_Modules/Locations/src/module/admin.ts
```diff
@@ -0,0 +1,172 @@
...
+    if (invalidLocation) {
+      throw new InvalidParamException(message ?? 'Invalid location request.');
+    }
+
+    const [ocpiLocation, citrineLocation] =
```
langiole, line 56: We may want to consider using the `core` naming prefix instead of `citrine`. 

E.g. `citrineLocation` -> `coreLocation`

00_Base/src/model/DTO/admin/AdminLocationDTO.ts
```diff
@@ -0,0 +1,153 @@
...
+        ? EVSE_INVALID_MSG
+        : connectorInvalid
+          ? CONNECTOR_INVALID_MSG
+          : '';
+
+    return [invalid, finalMessage];
+  };
+}
+
+export class AdminEVSEDTO {
```
langiole, line 126: I usually prefer the casing to be as: `AdminEvseDto`.

The all casing in this example seems a bit crazy cause the EVSE and DTO bleed into each other in a way.

00_Base/src/model/DTO/admin/AdminLocationDTO.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  @IsBoolean()
+  @Optional()
+  publish?: boolean;
+
+  @IsArray()
+  @Optional()
+  evses?: AdminEVSEDTO[];
+
+  static IS_LOCATION_INVALID = (
```
langiole, line 79: Shouldn't the validation annotations above take care of validation?

Why is this method needed?

00_Base/src/model/DTO/admin/AdminLocationDTO.ts
```diff
@@ -0,0 +1,153 @@
...
+  id?: number;
+
+  @IsString()
+  @MaxLength(2)
+  @Optional()
+  country_code!: string;
+
+  @IsString()
+  @MaxLength(3)
+  @Optional()
```
langiole, line 34: Why the optional annotation for party id, country code, and id?

Shouldn't these be required?

00_Base/src/util/OcpiLocationsUtil.ts
```diff
@@ -0,0 +1,65 @@
...
+            );
+
+          if (ocpiConnectorInfo) {
+            ocpiConnectorsMap[
+              `${TEMPORARY_CONNECTOR_ID(stationId, evseId, connectorId)}`
+            ] = ocpiConnectorInfo;
+          }
+        }
+
+        const ocpiEvseInfo = await this.ocpiEvseRepository.getEvseByEvseId(
```
langiole, line 51: This naming `ocpiEvseInfo` is a little bit confusing to me as it makes me think that there is a seperate `OcpiEvseInfo` object, but if I understand the code correctly this is just an `OcpiEvse`?

Can we not just name the variable as such?

00_Base/src/mapper/CitrineOcpiLocationMapper.ts
langiole, line 1: Ditto in this file about the naming of ocpi objects with the `info` postfix.

langiole, line 1: Is it fair to just name this file as LocationMapper? To match naming convention of CDR and Session mapper?

00_Base/src/mapper/CitrineOcpiLocationMapper.ts
```diff
@@ -105,9 +112,10 @@ export class CitrineOcpiLocationMapper implements IOcpiLocationMapper {
           ];
 
         if (!ocpiEvseInfo) {
-          throw new NotFoundException(
-            `OCPI EVSE ${UID_FORMAT(evseAttributes.station_id, evseAttributes.id)} does not exist.`,
+          this.logger.warn(
+            `OCPI EVSE ${UID_FORMAT(evseAttributes.station_id, evseAttributes.id)} does not exist - will skip.`,
```
langiole, line 116: Are all caps method names like `UID_FORMAT` usually a thing? 🤔 

I feel like that's typically for final vars only.

Also, I am questioning if this method should really live instead EvseDTO or if it is better suited for this mapper class.

Reason being that then we avoid having to put OCPI 2.0.1 references inside the EvseDTO.

00_Base/src/mapper/CitrineOcpiLocationMapper.ts
```diff
@@ -249,10 +259,10 @@ export class CitrineOcpiLocationMapper implements IOcpiLocationMapper {
     // TODO add remaining capabilities
     const capabilities: Capability[] = [];
 
-    if (authorizeRemoteStart === this.trueAttributeValue) {
+    if (authorizeRemoteStart === this.TRUE_ATTRIBUTE_VALUE) {
```
langiole, line 262: Maybe in the future we can instead define an enum `VariableAttributeValues`.

Then this can read as `VariableAttributeValues.TRUE`.

Only thing is that I don't know if this would work since the values might have different types.

03_Modules/Locations/src/module/admin.ts
```diff
@@ -0,0 +1,172 @@
...
+      this.mapAdminLocationDtoToEntities(adminLocationDto);
+    const savedCitrineLocation =
+      await this.locationRepository.createOrUpdateLocationWithChargingStations(
```
langiole, line 59: Could this method `createOrUpdateLocationWithChargingStations` not just be named `createOrUpdateLocation`?

03_Modules/Locations/src/module/admin.ts
```diff
@@ -0,0 +1,172 @@
...
+        citrineLocation,
+      );
+    ocpiLocation[OcpiLocationProps.citrineLocationId] = savedCitrineLocation.id;
```
langiole, line 62: Starting to question if this "Props" convention is a good move, as it seems to make readability a bit more difficult imo.

03_Modules/Locations/src/module/admin.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    const locationDto = this.locationMapper.mapToOcpiLocation(
+      savedCitrineLocation,
+      chargingStationVariableAttributesMap,
+      savedOcpiLocation,
+    );
+
+    if (broadcast) {
+      await this.locationsBroadcaster.broadcastOnLocationCreateOrUpdate(
```
langiole, line 99: Don't we have DB listeners that should emit a broadcast when the DB updates above are made?

03_Modules/Locations/src/module/admin.ts
```diff
@@ -0,0 +1,172 @@
...
+      for (const adminConnector of adminEvse.connectors ?? []) {
+        await this.ocpiConnectorRepository.createOrUpdateOcpiConnector(
+          this.mapAdminConnectorToOcpiConnector(adminConnector, adminEvse),
+        );
+      }
+    }
+
+    let chargingStationVariableAttributesMap = {};
+
+    if (savedCitrineLocation.chargingPool) {
```
langiole, line 81: So what happens if user specifies a full `AdminEvseDto` and `AdminConnectorDto` objects in the request?

I'm assuming any of the properties that are fulfilled by variable attributes are ignored?

langiole (approved):
Looks good!

I don't think I identified any critical changes that must be made. 

Mostly made comments on formatting.

00_Base/src/broadcaster/locations.broadcaster.ts
```diff
@@ -30,20 +32,24 @@ export class LocationsBroadcaster extends BaseBroadcaster {
...
+
+    const params = PutLocationParams.build(Number(locationDto.id), locationDto);
+
+    try {
+      await this.locationsClientApi.broadcastToClients(
+        locationDto.country_code,
+        locationDto.party_id,
+        ModuleId.Locations,
+        params,
+        this.locationsClientApi.putLocation.bind(this.locationsClientApi),
```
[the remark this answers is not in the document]
yuvenus: oh yeah it's a weird quirk, the binding essentially passes that instance of the locationsClientApi so the broadcaster can actually use it to invoke that passed in method :O i think there's another way to do this without the bind, but we've all settled on doing that in all our broadcast calls 😅

00_Base/src/repository/OcpiEvseRepository.ts
```diff
@@ -60,16 +60,17 @@ export class OcpiEvseRepository extends SequelizeRepository<OcpiEvse> {
...
-          lastUpdated: evse.lastUpdated,
-        },
-        savedOcpiEvse.id,
-      );
+      const values: Partial<OcpiEvse> = {};
+      values.physicalReference = evse.physicalReference ?? undefined;
+      values.removed = evse.removed !== undefined ? evse.removed : undefined;
+      values.lastUpdated = evse.lastUpdated ?? undefined;
+
+      await this._updateByKey({ ...values }, savedOcpiEvse.id);
```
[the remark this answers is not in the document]
yuvenus: i was back on forth on keeping it inline with spreads vs doing it as another variable, perhaps something we can talk about preference wise with the spreads 🤔 it was starting to get a bit unwieldy 🤣

00_Base/src/model/DTO/admin/AdminLocationDTO.ts
yuvenus: they're optional if the AdminLocationDTO is used in the "update" form rather than the "create" form :( i didn't play around much with the annotations though to see if there was a way to tell it "ignore the validation under these circumstances" -- let me see if there's a way to get that going, then i can remove that validation method 🤔

03_Modules/Locations/src/module/admin.ts
yuvenus: we don't currently have DB listeners that broadcast automatically bc we want to give some flexibility to the user so that they can choose to broadcast (sometimes you just want to edit something without telling the MSPs), regardless if a database change was made. it's kind of like a "force publish" trigger, but perhaps we can revisit with a better approach 🤔

00_Base/src/repository/OcpiLocationRepository.ts
```diff
@@ -77,6 +77,54 @@ export class OcpiLocationRepository extends SequelizeRepository<OcpiLocation> {
...
+    location: Partial<OcpiLocation>,
+  ): Promise<OcpiLocation> {
+    if (location[OcpiLocationProps.citrineLocationId]) {
+      const [savedOcpiLocation, ocpiLocationCreated] =
+        await this._readOrCreateByQuery({
+          where: {
+            [OcpiLocationProps.citrineLocationId]:
+              location[OcpiLocationProps.citrineLocationId],
+          },
+          defaults: {
```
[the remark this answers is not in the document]
yuvenus: yup exactly -- i'm using readOrCreateByQuery to have some shortcuts since we only want to create if it doesn't exist, otherwise just simply update the object.

00_Base/src/repository/OcpiLocationRepository.ts
```diff
@@ -77,6 +77,54 @@ export class OcpiLocationRepository extends SequelizeRepository<OcpiLocation> {
     );
   }
 
+  async createOrUpdateOcpiLocation(
+    location: Partial<OcpiLocation>,
+  ): Promise<OcpiLocation> {
+    if (location[OcpiLocationProps.citrineLocationId]) {
```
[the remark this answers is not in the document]
yuvenus: i was trying to do good by the ocpi location props we had, but agreed i think it might just be difficult to read as is 😬 i will update and hold off on the props for now...

03_Modules/Locations/src/module/admin.ts
yuvenus: i think they're good for some cases (like if you're constructing a query and you want to ensure the query has the right property), but maybe not so good for cases where you're just trying to grab that property from the object, since the IDE should be able to tell you if the property is real or not. 

i've updated general usages in this PR to follow those rules (only include for queries, don't include when doing object assignments/grabbing a property since we already know those properties are legit)

03_Modules/Locations/src/module/admin.ts
yuvenus: i only made this explicit in the event that someone just wants to update locations without charging stations involved (given that that's how most of our `createOrUpdate` methods are named and function, they don't necessarily touch the nested objects), but if that doesn't come to pass then we can update back to createOrUpdateLocation

00_Base/src/repository/OcpiLocationRepository.ts
```diff
@@ -77,6 +77,54 @@ export class OcpiLocationRepository extends SequelizeRepository<OcpiLocation> {
...
+            [OcpiLocationProps.publish]: location[OcpiLocationProps.publish],
+            [OcpiLocationProps.lastUpdated]:
+              location[OcpiLocationProps.lastUpdated],
+            [OcpiLocationProps.timeZone]: location[OcpiLocationProps.timeZone],
+          },
+        });
+      if (!ocpiLocationCreated) {
+        const values: Partial<OcpiLocation> = {};
+        values[OcpiLocationProps.citrineLocationId] =
+          location[OcpiLocationProps.citrineLocationId] ?? undefined;
```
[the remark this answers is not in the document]
yuvenus: updated, goodbye props!

00_Base/src/repository/OcpiLocationRepository.ts
```diff
@@ -77,6 +77,54 @@ export class OcpiLocationRepository extends SequelizeRepository<OcpiLocation> {
...
+        values[OcpiLocationProps.publish] =
+          location[OcpiLocationProps.publish] !== undefined
+            ? location[OcpiLocationProps.publish]
+            : undefined;
+        values[OcpiLocationProps.lastUpdated] =
+          location[OcpiLocationProps.lastUpdated] ?? undefined;
+        values[OcpiLocationProps.timeZone] =
+          location[OcpiLocationProps.timeZone] ?? undefined;
+
+        await this._updateByKey({ ...values }, savedOcpiLocation.id);
```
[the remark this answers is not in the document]
yuvenus: good point, updated!

03_Modules/Locations/package.json
```diff
@@ -29,6 +29,7 @@
     "@citrineos/base": "1.3.1",
     "@citrineos/util": "1.3.1",
     "deasync-promise": "^1.0.1",
-    "tslog": "^4.9.2"
+    "tslog": "^4.9.2",
+    "@types/geojson": "7946.0.14"
```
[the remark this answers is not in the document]
yuvenus: i didn't want to add a dependency, but that dang `Point`... :(

00_Base/src/model/DTO/admin/AdminLocationDTO.ts
langiole: In that case, maybe we should define a `CreateLocationRequest` and an `UpdateLocationRequest`? 

I guess the only issue there is that there might be a lot of redundancy.

00_Base/src/mapper/CitrineOcpiLocationMapper.ts
yuvenus: removed info! 👍

00_Base/src/util/OcpiLocationsUtil.ts
yuvenus: makes sense! done 👍

00_Base/src/mapper/CitrineOcpiLocationMapper.ts
yuvenus: yup this was made at a time where i thought we'd have more flexible location mapper (ocpi 3.0 mapper when?? 👀) but i think at the moment it just makes sense to simplify the naming until that time comes. updated!

00_Base/src/mapper/CitrineOcpiLocationMapper.ts
yuvenus: i was using this to indicate it's a static method but can change the naming convention based on what we decide 👍 the general placement of this method is indeed a bit strange... i think i'll move this into its own dedicated util or const file 🤔

03_Modules/Locations/src/module/api.ts
```diff
@@ -138,4 +154,24 @@ export class LocationsModuleApi
...
+   **/
+
+  // TODO add "admin endpoint" annotation for "regular" error handling when available
+  @Put('/admin')
+  @ResponseSchema(OcpiEmptyResponse, {
+    statusCode: HttpStatus.OK,
+    description: 'Successful response',
+  })
+  async createLocation(
+    @QueryParam('broadcast') broadcast: boolean,
```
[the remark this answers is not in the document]
yuvenus: it's required so that the endpoint validation can stop people from creating locations and wondering why their broadcasts aren't going to MSP if it silently defaults to false. 

if we're not worried this is a problem where people are wondering why the broadcasts aren't happening if they didn't set broadcast true, i can make this an optional field and default it to false 🤔

00_Base/src/util/OcpiLocationsUtil.ts
yuvenus: yup map could've worked here, i'm just used to javascript land where we can add straight to empty objects 😅 perhaps another thing we can decide on (always use maps or records or decide on situations where one is better than the other)

03_Modules/Locations/src/module/admin.ts
yuvenus: problem solved now that this method is removed thanks to the built in validation methods from our class validation annotations 😎

00_Base/src/model/DTO/admin/AdminLocationDTO.ts
yuvenus: makes sense! i think in some other places we have EVSE, but for this particular class it's been updated to `AdminEvseDTO` 👍

00_Base/src/model/DTO/admin/AdminLocationDTO.ts
yuvenus: aha, figured it out! you can add validation groups 😄 validation groups now added and they're properly preventing empty values only on create

00_Base/src/mapper/CitrineOcpiLocationMapper.ts
yuvenus: upon another look this is overkill, i think i'll just keep the string nice and simple 😄 reverted back to a simple string and removed this const. 

agreed maybe in the future we could use some sort of enum if the values become more varied 🤔

00_Base/src/broadcaster/locations.broadcaster.ts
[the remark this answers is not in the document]
elliot-sabitov: if you pass in `this.locationsClientApi.putLocation` alone without the bind - this is a function reference which will then later be invoked but the issue will be that the function alone will not have a context. To clarify, this JS function alone, will not be aware that it is a class function, nor will it be aware or anything related to the class like the class instance properties. Thus, in essense, any lines within that function that do `this.anything...` will all fail with `unable to read property of undefined - reading 'anything'` or something similar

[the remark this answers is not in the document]
elliot-sabitov: The `.bind()` ensures that the context is persisted - in fact, the not-so-new-anymore addition of the arrow function syntax `=>` actually was created in part to resolve this issue with binding ensuring that the result of the `() => {}` syntax is a bound function. I tried for a bit to write out an example to explain better but I think maybe just looking at other resources online, maybe something like this - https://dev.to/scooperdev/use-arrow-function-to-correctly-binds-this-1dnp, can help demystify this JS specific behavior in more detail, but I am also happy to jump on a call to try and clarify as well!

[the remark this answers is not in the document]
elliot-sabitov: As an alternative here, we can write it like this: 
```
await this.locationsClientApi.broadcastToClients(
      ocpiLocation ? ocpiLocation[OcpiLocationProps.countryCode] : 'US',
      ocpiLocation ? ocpiLocation[OcpiLocationProps.partyId] : 'CPO',
      ModuleId.Locations,
      params,
      () => {
        this.locationsClientApi.patchConnector(params)
      },
    );
```
or something similar and just adjust base to simply trigger the callback, but we may need to return await - but it is just another alternative - I felt like binding looked cleaner

PR #91, opened 2024-07-18
feat: Unregister Client from CPO and Delete Tenant endpoints

elliot-sabitov:
feat: adding delete tenant admin endpoint
feat: created new DELETE /unregister-client endpoint to remove a client from a given CPO
fix: noticed that @Middleware() annotation is read in via metadata and middleware is applied where it is not intended, so removed where it shouldn't be present since in most cases we apply middleware in the code via other annotations.
fix: adjusting logic to ensure that queries are properly handled within a single transaction by making sure that they are properly nested, because using `deleteAllByQuery` creates a new transaction internally and cannot be used in this way to associate the queries to the transaction intended to wrap the queries.
feat: adding logic to send a DELETE credentials request to the client when performing delete tenant and unregister tenant operations
chore: adjusting to use 127.0.0.1 instead of localhost because `typed-rest-client` seems to default to IPv6 when using localhost and resolution fails


lydiazcheng (commented):
The change looks good! 

But would be great if we add a sequalize transaction for 2 delete queries and even perfect is to have write lock on ClientInformation table so that we can make sure no new entity with the cpoTenantId can be created before we delete that CpoTenant.

elliot-sabitov:
> The change looks good!
> 
> It would be great if we add a sequalize transaction for 2 delete queries and even perfect is to have write lock on ClientInformation table so that we can make sure no new entity with the cpoTenantId can be created before we delete that CpoTenant.

omg 😮‍💨 I remember thinking to my self - "need to wrap in transaction" but forgot - should be there now 👍

lydiazcheng (approved):

03_Modules/Credentials/src/module/api.ts
```diff
@@ -172,4 +173,24 @@ export class CredentialsModuleApi
...
+    statusCode: HttpStatus.OK,
+    description: 'Successful response',
+    examples: {
+      success: {
+        summary: 'A successful response',
+        value: MOCK_EMPTY,
+      },
+    },
+  })
+  async deleteTenant(
```
thanaParis, line 188: Is there a separate PR with an endpoint that allows you to deregister a specific MSP rather than delete the whole CPO tenant? I'm not seeing one--deleting a whole CPO tenant is a nice extra feature but what we really needed is a way to make a specific CPO call DELETE for a specific MSP, then remove that MSP's access to that CPO

thanaParis (changes requested):
Delete by tenant is insufficient for OCPI support.

03_Modules/Credentials/src/module/api.ts
elliot-sabitov: I see, I think I must have mistunderstood. I will leave this endpoint and add an endpoint to unregister one MSP so that a given CPO can use this endpoint to unregister an MSP.

elliot-sabitov: PR updated with the new `/unregister-client` endpoint

00_Base/src/services/credentials.service.ts
```diff
@@ -474,6 +474,85 @@ export class CredentialsService {
...
+      );
+    }
+
+    for (const clientInformation of clientInformationMatches) {
+      const credentialsRoleMatches = clientInformation[ClientInformationProps.clientCredentialsRoles]
+        .filter((clientInformations) => {
+          return clientInformations.country_code === clientCountryCode && clientInformations.party_id === clientPartyId;
+        });
+      for (let credentialsRoleMatch of credentialsRoleMatches) {
+        await credentialsRoleMatch.destroy();
```
thanaParis, line 551: This removes the credentials on our side, but fails to call DELETE on the client. It is necessary to call DELETE on the client before removing the credentials on our side.

elliot-sabitov: Sorry for missing that 🤦 I will add that - also, I presume that for DELETE tenant, we would trigger same behavior for each client tied to that tenant as well, so I will add that in too, but correct me if I am wrong 🙏

00_Base/src/services/credentials.service.ts
```diff
@@ -474,6 +474,85 @@ export class CredentialsService {
...
+    const cpoTenant = await this.cpoTenantRepository.readOnlyOneByQuery({
+      where: {
+        id: tenantId,
+      },
+    });
+    if (!cpoTenant) {
+      throw new NotFoundError('CpoTenant not found');
+    }
+
+    const transaction =
```
ChrisWeissmann, line 487: This is missing the association with the query.

00_Base/src/services/credentials.service.ts
elliot-sabitov: implemented above

00_Base/src/model/UnregisterClientRequestDTO.ts
```diff
@@ -0,0 +1,28 @@
+import { Column, DataType } from 'sequelize-typescript';
+import { IsNotEmpty, IsString, Length } from 'class-validator';
+
+export class UnregisterClientRequestDTO {
+  @Column(DataType.STRING(3))
```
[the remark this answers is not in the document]
elliot-sabitov: adjusted 👍

00_Base/src/services/credentials.service.ts
elliot-sabitov: adjusted 👍 🙇

yuvenus (approved):

thanaParis (approved):

PR #96, opened 2024-07-22
Feature: Created Transactions with Country Code and Party Id Postgres View.

elliot-sabitov:
feat: created ViewTransactionsWithPartyIdAndCountryCode model and view to allow us to query for transactions for particular partyId and country code. Created a migration to create the view. Adjusted sequelize initialization to drop and recreate views before and after running the sequelize.sync to ensure that migrations do not have a conflict when trying to alter a column for a table that is part of a view. Created a repository and a simple test (commented out) to show that using this view we are able to query for transactions based on country code and party id.

Screenshot 2024-07-22 at 3 18 17 PM
Screenshot 2024-07-22 at 2 58 51 PM

Pairs with: https://github.com/citrineos/citrineos-core/pull/176

### DIFF
--- a/00_Base/src/index.ts
+++ b/00_Base/src/index.ts
@@ -29,6 +29,7 @@ import {
 import { SessionBroadcaster } from './broadcaster/session.broadcaster';
 import { CdrBroadcaster } from './broadcaster/cdr.broadcaster';
 
+export { createViewTransactionsWithPartyIdAndCountryCodeSql, dropViewTransactionsWithPartyIdAndCountryCodeSql } from './sql/ViewTransactionsWithPartyIdAndCountryCode';
 export { NotFoundException } from './exception/NotFoundException';
 export { FunctionalEndpointParams } from './util/decorators/FunctionEndpointParams';
 export { PaginatedOcpiParams } from './trigger/param/paginated.ocpi.params';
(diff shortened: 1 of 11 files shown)


elliot-sabitov:
We found that `VIEW`s don't store data, but instead still execute a complex query essentially ensuring a cleaner interface to large queries. There are `MATERIALIZED VIEW`s that can be created to store the data, but they do not automatically update, and a refresh would need to be triggered manually. There would have to be other triggers set up either in code or the DB (or elsewhere) to trigger refresh of the data in the materialized view for this to work properly, but in addition, Postgres does not natively support `MATERIALIZED VIEW`s and there is only a work around which I am sure is akin to creating a "View Table" and updating it via DB triggers and rules. 

Either way though, since `VIEW`s do not directly improve performance, it becomes more of a nicety to keep the code clean, and since it is a brittle set up to get views working with our sequelize typescript setup, I think there isnt much point in using something like this since it sort of hacks support for views together since sequelize-typescript doesnt offer native support for views. I will close this PR for now. We can still consider using this if there are really super ugly queries that we want to move to a view, and I think this PR can be closed and can be used as a reference in those cases.

If we do see any use cases where `MATERIALIZED VIEW`s can be a good fit, we should keep that in mind and utilize this tool where appropriate. It could be very useful to have a precomputed `VIEW` that has the data needed to ensure optimal query performance. The caveats for updating / refreshing the `MATERIALIZED VIEW` and the workaround to get it to work with `sequelize` presented in this PR will both need to be accounted for, so we can use this PR as a reference in those cases as well.

PR #97, opened 2024-07-22
Feature - Real Time Authorization

elliot-sabitov:
feat: initializing RealTimeAuthorizer in Container in Tokens module and adjusting Server such that OCPI is initialized prior to System allowing us to get the RealTimeAuthorizer from Container and pass it to the EVDriver module constructor
feat: created a mock tokens authorize endpoint in mock-client to send a mock response for testing
feat: ensuring request is routed properly using the already available CpoTenant


03_Modules/Tokens/src/module/authorizer/RealTimeAuthorizer.ts
```diff
@@ -0,0 +1,126 @@
...
+    const idToken = authorization.idToken as IdToken;
+    const idTokenId = idToken.id;
+    const ocpiToken = await this.tokensRepository.getOcpiTokenByAuthorizationId(authorization.id);
+    if (!ocpiToken) {
+      throw new BadRequestError(`Could not find OCPI token with id ${idTokenId}.`);
+    }
+    return ocpiToken;
+  }
+
+  private async getCpoTenant(ocpiToken: OcpiToken, context: IMessageContext): Promise<CpoTenant> {
```
yuvenus, line 60: minor, seems like there's a linter warning here -- `'context' is defined but never used. Allowed unused args must match /^_/u`

03_Modules/Tokens/src/module/authorizer/RealTimeAuthorizer.ts
```diff
@@ -0,0 +1,126 @@
...
+      if (authorizationInfo.allowed === AuthorizationInfoAllowed.Allowed) {
+        result.status = AuthorizationStatusEnumType.Accepted;
+      }
+    } catch (e: any) {
+      this.logger.error('Error: could not authorize token -' + e.message);
+    }
+    return result;
+  }
+
+  private async getOcpiToken(authorization: Authorization, context: IMessageContext): Promise<OcpiToken> {
```
yuvenus, line 50: minor, seems like there's a linter warning here -- 'context' is defined but never used. Allowed unused args must match /^_/u

mock-client/tokens.ts
```diff
@@ -43,4 +49,32 @@ export class TokensController extends BaseController {
...
+  @Post('/:tokenId/authorize')
+  @ResponseSchema(PaginatedTokenResponse, {
+    statusCode: HttpStatus.OK,
+    description: 'Successful response',
+    examples: {
+      success: TOKENS_LIST_MOCK,
+    },
+  })
+  async authorizeToken(): Promise<AuthorizationInfoResponse> {
+    const ocpiToken = new OcpiToken();
```
yuvenus, line 62: will there be a todo to make these values dynamic or is this just a testing endpoint that can be removed? :O

yuvenus (approved):
looks good! just a couple of linter errors 👍

mock-client/tokens.ts
elliot-sabitov: this is one of the mock endpoints that we use for testing... I am not sure that we want to remove it, I think it can be helpful in testing, but yes, we can technically just have an OpenAPI spec and use `mockoon` to mock it, but since they havent published OpenAPI for v2.2.1, this allows us to write a controller that will kill two birds in one stone in that it helps generate the spec while also performing the mock... You can even layer in `mockoon` to then run using the spec that this `mock-client` produces... I suppose another way of doing it would be to just write out the `json` spec 🤔

03_Modules/Tokens/src/module/authorizer/RealTimeAuthorizer.ts
elliot-sabitov: fixed

03_Modules/Tokens/src/module/authorizer/RealTimeAuthorizer.ts
elliot-sabitov: fixed

03_Modules/Tokens/src/module/authorizer/RealTimeAuthorizer.ts
```diff
@@ -0,0 +1,154 @@
...
+  OcpiToken,
+  PostTokenParams,
+  ServerCredentialsRoleProps,
+  TokensClientApi,
+  TokensRepository,
+  UnsuccessfulRequestException,
+} from '@citrineos/ocpi-base';
+import { BadRequestError } from 'routing-controllers';
+import { ILogObj, Logger } from 'tslog';
+import { AuthorizationInfo } from '@citrineos/ocpi-base/dist/model/AuthorizationInfo';
```
langiole, line 24: Incorrect import.

03_Modules/Tokens/src/module/authorizer/RealTimeAuthorizer.ts
langiole, line 1: My recommendation would be to move this file into base for the sake of consistency.

We can move it back into Tokens once the effort is made to move all other module related files into their respective module.

PR #100, opened 2024-07-24
OCPI Tokens PATCH edit

danielyoo227:
Editing Tokens PATCH in OCPI to support update OCPP Authorization during the flow. Also refactored to move common logic for readability. Added transaction support for the update logic so that all updates (Authorization, Token) run in a transaction to avoid side effects.

### DIFF
--- a/00_Base/src/repository/TokensRepository.ts
+++ b/00_Base/src/repository/TokensRepository.ts
@@ -230,20 +230,22 @@ export class TokensRepository extends SequelizeRepository<OcpiToken> {
   }
 
   private async updateOcpiToken(existingOcpiToken: OcpiToken, partialToken: Partial<OcpiToken>, transaction: SequelizeTransaction): Promise<OcpiToken> {
-    const [updatedCount, [updatedToken]] = await OcpiToken.update(partialToken, {
+    const [updatedCount, updatedTokens] = await OcpiToken.update(partialToken, {
       where: {id: existingOcpiToken.id},
       returning: true,
       transaction
     });
 
-    if (updatedCount > 1) {
-      this.logger.warn(`More than one token updated for primary ID ${existingOcpiToken.id}`);
-    }
-
-    if (!updatedToken) {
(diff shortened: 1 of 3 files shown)


00_Base/src/repository/TokensRepository.ts
```diff
@@ -245,50 +168,110 @@ export class TokensRepository extends SequelizeRepository<OcpiToken> {
...
+        },
+        country_code: queryParams.country_code,
+        party_id: queryParams.party_id,
+        type: queryParams.type,
+      },
+    });
+  }
+
+  private async updateOcpiToken(existingOcpiToken: OcpiToken, partialToken: Partial<OcpiToken>, transaction: SequelizeTransaction): Promise<OcpiToken> {
+    const [updatedCount, [updatedToken]] = await OcpiToken.update(partialToken, {
```
yuvenus, line 233: should this be `const [updatedCount, updatedToken]` instead so that `updatedToken` is not in a list, or is the list intentional?

yuvenus (approved):
looks good! did you also have a chance to run the linter, it seems like there could be some formatting stuff that would be picked up but i'm not 100% sure 🤔

00_Base/src/repository/TokensRepository.ts
danielyoo227: I meant to always take the first token in the update but good point, I think this can NPE right? Will change!

langiole:
@danielyoo227 I am assuming when you tested this you passed in full request body instead of partial?

Reason I ask is because if you pass in a partial object you will get validation errors since `TokenDTO` is used in the declaration.

We will need to update this request body use something other than `TokenDTO` since all fields are optional for PATCH.

According to the spec:
> 12.2.1.3. PATCH Method
> Same as the PUT method, but only the fields/objects that have to be updated have to be present, other fields/objects that are not specified are considered unchanged.
> 
> Any request to the PATCH method SHALL contain the last_updated field.

PR #109, opened 2024-07-26
chore: version bump

elliot-sabitov:


danielyoo227 (approved):

00_Base/package.json
```diff
@@ -32,8 +32,8 @@
     "@types/lodash.startcase": "^4.4.9"
   },
   "dependencies": {
-    "@citrineos/base": "1.3.1",
-    "@citrineos/data": "1.3.1",
+    "@citrineos/base": "1.3.2",
+    "@citrineos/data": "1.3.2",
```
ChrisWeissmann, line 36: won't this still break, since 1.3.2 is not on npm yet?

ChrisWeissmann (commented):
I think having a PR that just updates the actual version would be cleaner instead of 2 things at once.

00_Base/package.json
elliot-sabitov: I adjusted our workflow to git clone a specific branch because otherwise it fails the other way where it pulls `ocpi` main which doesnt have new code for example the RealTimeAuthorizer, and then fails

PR #111, opened 2024-07-29
Adding retry setup support for sequelize

danielyoo227:
Added retry support for sequelize. Refactored code to make server setup to support async.


elliot-sabitov:
This is super nice - we may want to create some tickets here because we will probably want to do the same thing with `core`s `DefaultSequelizeInstance` so perhaps some of this logic can instead be moved to `base` in `core` or something... In addition, I saw similar stuff with `amqp` if I did not have docker running, and server crashes, but we may want to retry potentially. 👍

PR #114, opened 2024-07-30
(Release 1.4) Tokens datasource

langiole:


00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,226 @@
+import {ITokensDatasource} from "./ITokensDatasource";
+import {TokenDTO} from "../model/DTO/TokenDTO";
+import {Authorization, SequelizeAuthorizationRepository} from "@citrineos/data";
+import {OcpiToken, SingleTokenRequest} from "../model/OcpiToken";
+import {OcpiTokensMapper} from "../mapper/OcpiTokensMapper";
+import {TokenType} from "../model/TokenType";
+import {Op} from "sequelize";
+import {OcpiLogger} from "../util/logger";
+import {TokensRepository} from "../repository/TokensRepository";
+import {IdTokenEnumType} from "../../../../citrineos-core/00_Base";
```
yuvenus, line 10: fix import

00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,226 @@
...
+
+        const updatedToken = await this.tokensRepository.updateByKey(
+            partialToken,
+            existingOcpiToken.id,
+        );
+
+        if (!updatedToken) {
+            throw new UnknownTokenException('Token not found in the database');
+        }
+        //
```
yuvenus, line 104: if the commented out code is no longer needed, can it be removed?

00_Base/src/mapper/cdr.mapper.ts
```diff
@@ -18,14 +18,15 @@ import { ILogObj, Logger } from 'tslog';
 import { OcpiLocationRepository } from '../repository/OcpiLocationRepository';
 import { TokensRepository } from '../repository/TokensRepository';
```
yuvenus, line 19: minor linter warning: TokensRepository isn't used, you can get rid of it

00_Base/src/repository/TokensRepository.ts
```diff
@@ -38,224 +37,6 @@ export class TokensRepository extends SequelizeRepository<OcpiToken> {
     );
   }
```
yuvenus, line 38: seems that there are a bunch of linter warnings for unused imports, seems that you can get rid of a bunch of the imports then 👍

yuvenus (approved):
looks good to me! just a few linter warnings and what not, run lint-fix and we should be good 👍

00_Base/src/datasources/TokensDatasource.ts
langiole: There is a ticket to fix Tokens patch.

You can see that the return of this method is just an empty Token DTO.

Was thinking just leave this in until we revisit it.

00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,229 @@
...
+    const updatedToken = await this.tokensRepository.updateByKey(
+      partialToken,
+      existingOcpiToken.id,
+    );
+
+    if (!updatedToken) {
+      throw new UnknownTokenException('Token not found in the database');
+    }
+    //
+    // const newOCPPAuth =
```
ChrisWeissmann, line 108: Either use it or delete it. Commenting out code usually just makes things harder to understand ;)

00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,229 @@
...
+    //       idToken: partialToken.uid!,
+    //       type: OCPITokensMapper.mapOcpiTokenTypeToOcppIdTokenType(partialToken.type),
+    //     },
+    //   );
+    //
+    // if (!updatedOcppAuth) {
+    //   throw new Error('Could not save token');
+    // }
+
+    return new TokenDTO();
```
ChrisWeissmann, line 124: Doesn't this just return a new empty object? Doesn't seem to be a very useful response?

00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,229 @@
...
+    } catch (error) {
+      this.logger.error('Error saving token', error);
+      throw error;
+    }
+  }
+
+  async updateBatchedTokens(tokens: TokenDTO[]): Promise<void> {
+    const batchFailedTokens: TokenDTO[] = [];
+
+    // TODO: do an actual bulk update
```
ChrisWeissmann, line 162: 1. I am not convinced this should be a bulk update. having it in a try catch loop seems more resilient to failures of individual bad tokens. This way we can investigate those cases and still have most updates go through.
2. Can we make it a Ticket instead of a code TODO if we decide it's worth it.

00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,229 @@
...
+    type: IdTokenEnumType,
+  ): Promise<TokenDTO | undefined> {
+    const ocppAuth =
+      await this.authorizationRepository.readOnlyOneByQuerystring({
+        idToken,
+        type,
+      });
+
+    if (!ocppAuth) {
+      // TODO better error
```
ChrisWeissmann, line 190: Can we fill out this todo now?

00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,229 @@
...
+      return undefined;
+    }
+
+    const ocpiToken = await this.getOcpiTokenByAuthorizationId(ocppAuth.id);
+
+    if (!ocpiToken) {
+      // TODO better error
```
ChrisWeissmann, line 197: Can we fill out this todo now?

00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,229 @@
...
+  async getOcpiTokenByAuthorizationId(
+    authorizationId: string,
+  ): Promise<OcpiToken | undefined> {
+    const ocpiToken = await this.tokensRepository.readOnlyOneByQuery({
+      where: {
+        authorization_id: authorizationId,
+      },
+    });
+    if (!ocpiToken) {
+      // TODO better error
```
ChrisWeissmann, line 213: Can we fill out this todo now?

00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,229 @@
...
+      return Promise.resolve(undefined);
+    }
+    return Promise.resolve(ocpiToken);
+  }
+
+  getTokens(): Promise<TokenDTO[]> {
```
ChrisWeissmann, line 219: Is this relevant? It doesn't seem to be called anywhere and the implementation doesn't seem to do much. If you really want to keep it, I think the response should be an error that the method isn't implemented yet. So that the next person doesn't rely on using it.

00_Base/src/datasources/TokensDatasource.ts
langiole: Removed for now

00_Base/src/datasources/TokensDatasource.ts
langiole: Updated to not use batch

00_Base/src/datasources/TokensDatasource.ts
langiole: Removed

00_Base/src/datasources/TokensDatasource.ts
```diff
@@ -0,0 +1,316 @@
...
+        },
+        defaults: {
+          ...baseTokenData,
+          authorization_id: authorizationId,
+          country_code: tokenDto.country_code,
+          party_id: tokenDto.party_id,
+        },
+      });
+
+    // TODO confirm all updatable properties
```
langiole, line 266: @ChrisWeissmann correct me if wrong but I think this TODO can be removed as all props are updatable.

00_Base/src/datasources/TokensDatasource.ts
langiole: Lmk what you think but I kinda like returning undefined here instead of throwing.

I think throwing makes sense when you are trying to patch a token that does not exist.

ChrisWeissmann: I agree, get is ok to return undefined, patch can throw as the operation is erroring. So let's remove the `TODO` here :)

00_Base/src/datasources/TokensDatasource.ts
ChrisWeissmann: Works for me.

00_Base/src/services/TokensAdminService.ts
```diff
@@ -108,7 +111,7 @@ export class TokensAdminService {
         if (
           response.status_code === OcpiResponseStatusCode.GenericSuccessCode
         ) {
-          await this.tokenRepository.updateBatchedTokens(response.data);
+          this.updateTokens(response.data);
```
ChrisWeissmann, line 114: Was the removal of the `await` on purpose? It's still an async call right.

langiole: It was not on purpose. Added back in.

ChrisWeissmann (approved):
Looks good to me! 🚀

PR #122, opened 2024-07-31
Fixing Patch logic for Tokens

danielyoo227:
Fixing PATCH logic. Also added reload logic for the GET flow since we need IdTokenInfo's information to map some of the TokenDTO data.

### DIFF
--- a/00_Base/src/repository/TokensRepository.ts
+++ b/00_Base/src/repository/TokensRepository.ts
@@ -8,13 +8,15 @@ import { OcpiToken, SingleTokenRequest } from '../model/OcpiToken';
 import { OcpiSequelizeInstance } from '../util/sequelize';
 import {
   Authorization,
+  IdToken,
+  IdTokenInfo,
   SequelizeAuthorizationRepository,
   SequelizeRepository,
   SequelizeTransaction,
 } from '@citrineos/data';
 import { OcpiServerConfig } from '../config/ocpi.server.config';
 import { OcpiLogger } from '../util/logger';
-import { IdTokenEnumType, SystemConfig } from '@citrineos/base';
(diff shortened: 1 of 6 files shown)


00_Base/src/repository/TokensRepository.ts
```diff
@@ -285,15 +312,79 @@ export class TokensRepository extends SequelizeRepository<OcpiToken> {
...
+      transaction,
+    });
+  }
+
+  private mapTokenDTOToPartialAuthorization(
+    existingAuth: Authorization,
+    tokenDTO: Partial<TokenDTO>,
+  ): Partial<IdTokenInfoType> {
+    const idTokenInfo: IdTokenInfoType = {
+      status: existingAuth.idTokenInfo?.status!,
```
yuvenus, line 365: linter error her, seems that it doesn't like the not null assertion here 🤔 i think it's solvable by making the type of id token into a `Partial<IdTokenInfoType>` instead...?

yuvenus (approved):
approved! seems there's a linter error that needs to be resolved before this is mergeable though 🤔

00_Base/src/repository/TokensRepository.ts
```diff
@@ -285,15 +312,79 @@ export class TokensRepository extends SequelizeRepository<OcpiToken> {
...
+      });
+    }
+
+    return await existingAuth.reload({
+      include: [{ model: IdTokenInfo }],
+      transaction,
+    });
+  }
+
+  private mapTokenDTOToPartialAuthorization(
```
langiole, line 360: Maybe we can put this method in the tokens mapper file?

langiole (approved):

00_Base/src/repository/TokensRepository.ts
danielyoo227: Yup, makes sense. Moved!

PR #123, opened 2024-07-31
(Release 1.4) Update NPM scripts to use cross-env where possible

yuvenus:
Core PR: https://github.com/citrineos/citrineos-core/pull/195

This PR adds the cross-env dependency (https://www.npmjs.com/package/cross-env) to allow cross-OS usability of the same NPM scripts that need to set environment variables. Only start was updated, but if there are more commands I should update, let me know!


ChrisWeissmann (approved):
I approve of this message!

PR #126, opened 2024-07-31
Feat/real time auth allowed offline

elliot-sabitov:
feat: adjusted logic such that ALLOWED_OFFLINE whitelist type will return Accepted status if there are any errors in communicating with the MSP
fix: issue with sequelize Model being initialized unintentionally, so instead just returning a simple object


yuvenus (approved):

danielyoo227 (approved):

thanaParis (approved):
Approved

PR #127, opened 2024-07-31
Feature/docker ocpi adjustments

elliot-sabitov:
- feat: moving latest core with ocpi server implementation into ocpi /Server and adjusting paths, npm run commands, docker file, etc. to use locally referenced `core` from within `ocpi`
- chore: for now copied most of /Server from core into ocpi because it closest matched what need to be able to run core with ocpi. This is a temporary solution and there will be a fast follow to instead move this to the demo repo.
- chore: moving Server to demo
- chore: adjusting npm run scripts to no longer start any server, but only be able to install, lint and build, which required workspace and dependency adjustments
- chore: cleanup unneeded Server_OLD, Swarm, etc references that were never used.
- chore: removing Server from lint
- chore: fixing dependency versioning issues and issues where errors came up because multiple versions of dependencies were installed because core still has the carrot ^ in the dependencies but OCPI had static versions
- chore: adjusting sequelize config to include the new maxRetries and retryDelay variables
- chore: fix build issues
- fix: adjusting OcpiServer initialization logic to ensure that modules are initialized in the constructor because they require deasync which cannot be ran from within a promise based function without seemingly running into an endless loop where the promise never resolves and system appears to hang.
- chore: re-aligning dependencies to make sure that the dependencies are properly hoisted and that there are no conflicts between different versions of dependencies being installed throughout the modules causing to build errors
- chore: removing --workspaces to ensure that all dev dependencies are installed along with the workspace dependencies otherwise some dependencies are missed.
- chore: simplifying prettier command


00_Base/src/config/sub/ServerConfigUtilMessageBrokerAmqp.ts
```diff
@@ -0,0 +1,19 @@
...
+export class ServerConfigUtilMessageBrokerAmqp {
+  @IsString()
+  @IsNotEmpty()
+  url: string;
+
+  @IsString()
+  @IsNotEmpty()
+  exchange: string;
+
+  constructor() {
```
elliot-sabitov, line 15: Please note: constructors in these `config` related classes are intended only to be defaults giving us the option to be able to eliminate the `local` version of the config which can be chosen as the "default"

thanaParis (changes requested):
Please re-add Server. There should be a way to run core + ocpi 2.2.1 from this repo.
There are a lot of changes in this PR. Please break it down into separate PRs where possible.

elliot-sabitov:
Broke this original PR down into smaller PRs:
- https://github.com/citrineos/citrineos-ocpi/pull/130
- https://github.com/citrineos/citrineos-ocpi/pull/129
- https://github.com/citrineos/citrineos-ocpi/pull/131

yuvenus (approved):

thanaParis (approved):
Approved!

PR #129, opened 2024-08-05
feat: renaming OcpiServerConfig to ServerConfig, breaking it down int…

elliot-sabitov:
feat: renaming OcpiServerConfig to ServerConfig, breaking it down into multiple files and creating some additional models to better match config with core, supporting defaults, to ensure a cleaner integration while allowing OCPI to layer additional configs.


00_Base/src/index.ts
```diff
@@ -77,7 +78,9 @@ export {
 } from './model/ClientInformation';
 export { ClientCredentialsRole } from './model/ClientCredentialsRole';
 export { fromCredentialsRoleDTO } from './model/ClientCredentialsRole';
-export { OcpiServerConfig } from './config/ocpi.server.config';
+export { ServerConfig } from './config/ServerConfig';
+export * from './config/sub';
```
ChrisWeissmann, line 82: Why export everything from sub? Maybe we should have single setup for ENUMS, I think that's what is actually needed on the outside.

00_Base/src/util/util.ts
```diff
@@ -14,9 +14,13 @@ export enum CountryCode {
...
 
-export const plainToClass = <T>(constructor: Constructable<T>, plain: T): T =>
+export const plainToClass = <T>(
+  constructor: Constructable<T>,
+  plain: T,
+  excludeExtraneousValues = true,
+): T =>
   plainToInstance(constructor, plain as T, {
-    excludeExtraneousValues: true,
+    excludeExtraneousValues,
```
ChrisWeissmann, line 23: Why this functional change?

ChrisWeissmann (commented):
What about the `/Server`. It still seems to do `this.config as OcpiServerConfig`

00_Base/src/config/ServerConfig.ts
```diff
@@ -0,0 +1,66 @@
+import { Service } from 'typedi';
+import { Enum } from '../util/decorators/enum';
+import { IsInt, IsNotEmpty, IsPositive, ValidateNested } from 'class-validator';
+import { Type } from 'class-transformer';
+import { SequelizeConfig } from './sub/SequelizeConfig';
+import { Env } from './sub/Env';
+import { ServerConfigData } from './sub/ServerConfigData';
```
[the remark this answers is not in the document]
elliot-sabitov: adjusted

00_Base/src/util/util.ts
elliot-sabitov: this helper method was hardly used, and it was previously always setting `excludeExtraneousValues` to true, but that will not parse child classes so I adjusted this to be a variable so we can default to true but are able to pass in `false` where needed

00_Base/src/index.ts
elliot-sabitov: The intent is to be able to import the sub classes as parts of the config into other classes where those config values are needed. You are correct in that this can be removed without any impact

ChrisWeissmann (approved):

langiole:
I think this PR might've broken the build for ocpi.

Can somebody else confirm?

00_Base/src/config/ocpi.server.config.ts
```diff
@@ -1,209 +0,0 @@
...
-  @IsOptional()
-  sync?: boolean;
-
-  @IsBoolean()
-  @IsOptional()
-  alter?: boolean;
-
-  @IsNumber()
-  @IsOptional()
-  maxRetries?: number;
```
langiole, line 47: Seems like this and `retryDelay` never made it to any other config.

Causing build errors on my end.

```
00_Base/src/util/sequelize.ts:71:51 - error TS2339: Property 'maxRetries' does not exist on type 'SequelizeConfig'.

71     const maxRetries = this.config.data.sequelize.maxRetries ?? 5;
                                                     ~~~~~~~~~~

00_Base/src/util/sequelize.ts:72:51 - error TS2339: Property 'retryDelay' does not exist on type 'SequelizeConfig'.

72     const retryDelay = this.config.data.sequelize.retryDelay ?? 5000;
```

PR #130, opened 2024-08-05
chore: created npm run clear and fresh adjustments

elliot-sabitov:
- chore: created npm run clear cache and enhancing npm run fresh to also clear cache after noticing issues after clearing local cache, clearing cache with fresh should make sure that any future issues are captured and resolved early.
- chore: adding missing dependencies to resolve build errors
- chore: removing core dependencies from modules that import ocpi-base since ocpi-base already import core dependencies.
- chore: adjusting npm run scripts to no longer start any server, but only be able to install, lint and build, which required workspace and dependency adjustments
- chore: removing no longer needed fresh, compile and clean commands form within modules since they will not be run individually, but will be built as part of the main bundle


00_Base/package.json
```diff
@@ -22,35 +19,44 @@
   "author": "S44",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@types/koa-logger": "^3.1.5",
```
ChrisWeissmann, line 22: I am confused about the dependency changes here. Why are they needed? 
Like why do we need koa-logger now? What changed that we now need sequelize?

elliot-sabitov: I am not sure why @types/koa-logger was not included, because `koa-logger` is in the dependencies, but for some reason the `@types/koa-logger` was missing. I am not sure how that happened. The other dependencies though, like sequelize, and others, seem to be needed otherwise there is an error thrown because it does not find these dependencies... 

I think our issue here is because of dependency hoisting. I found this article - https://www.jonathancreamer.com/inside-the-pain-of-monorepos-and-hoisting/ which explains it in more detail, but essentially, node will automatically hoist common dependencies to the root but if anything is misconfigured that may not work. I will look to see if we can remove these, but we may be limited in that we don't have a static package-lock.json that we use, and once we have those, it may be helpful...

elliot-sabitov: I looked at it a bit more, I think it is possible that there may have been something with `npm cache` or perhaps running `npm i` vs `npm run install-all` possibly caused the issue. When running `npm i` the `PATH` and other env vars may be different and I saw that my PATH when running `npm i` simply was my ENV PATH but running `npm run install-all` included other `node_modules` folders in the `PATH`... 

It seems that it depends on how we use the package dependencies, if we have Package A which has a dependency of Package B which then has a dependency of Package C, and we use Package C directly in Package A, then we have to include it in package.json in Package A. If Package A only interracts with anything in Package C through package B, for example,
```
import * as sequelize from '@citrineos/data';
```
or similar, then Package C does not need to be included in Package A...

Screenshot 2024-08-12 at 9 36 44 AM

ChrisWeissmann (approved):

PR #134, opened 2024-08-12
chore: error handling for POST /register-credentials-token-a endpoint

elliot-sabitov:
chore: adjusting exception to include error message when getting version details and resolving unhandled exception error.
fix: added additional error handling and applying the new @AsAdminEndpoint to ensure that the errors are caught
chore: updating temp serverVersionUrl to plugfest-eu variant instead of dallas


00_Base/src/services/credentials.service.ts
```diff
@@ -410,7 +411,7 @@ export class CredentialsService {
     const serverVersion = serverVersionResponse[0];
     // TODO, should version url should be in DB?
     const serverVersionUrl =
-      'https://plugfest-dallas.demo.citrineos.app:445/ocpi/versions';
+      'https://plugfest.demo.citrineos.app:445/ocpi/versions';
```
lydiazcheng, line 414: If you want, maybe we can temp add a query param for this serverVersionUrl besides this default value so that we are able to change it if necessary.

00_Base/src/services/credentials.service.ts
```diff
@@ -498,21 +499,32 @@ export class CredentialsService {
 
     const clientInformation = clientCpoTenant.clientInformation[0];
 
-    // await clientInformation.save(); // todo
-    await clientCpoTenant.save();
+    try {
+      await clientCpoTenant.save();
```
lydiazcheng, line 503: In the future, maybe we can have a improvement for this endpoint that not always create a new cpo tenant but only update if it existing. And also not store this temp credentials A but only formal credentials C.

03_Modules/Credentials/src/module/api.ts
```diff
@@ -162,6 +162,7 @@ export class CredentialsModuleApi
    * and register token B and C.
    */
   @Post('/register-credentials-token-a')
+  @AsAdminEndpoint()
```
lydiazcheng, line 165: Nice catch! Would you mind helping add this annotation for `@Post('/generate-credentials-token-a')` and `@Put('/regenerate-credentials-token')` as well?

lydiazcheng (approved):

00_Base/src/services/credentials.service.ts
elliot-sabitov: Agreed, once we have multi tenant support, this would be adjusted to not create a new tenant but instead update if one already exists... I think we do need to store the temp credentials though because MSP then uses that token to authenticate with the CPO when making the requests

03_Modules/Credentials/src/module/api.ts
elliot-sabitov: added

00_Base/src/services/credentials.service.ts
lydiazcheng: Please correct me if my understanding is wrong. I don't think CPO need to store token A because this temp token A is generated by MSP. CPO use this token A to authenticate with MSP and get its version and version details and then send CPO's token B to MSP. Then MSP use token B to authenticate with CPO and then send CPO its formal token C.  I think even though we call this endpoint as register token A, it finally register and store the formal token C.

elliot-sabitov: Ah yes, regarding Token A, it does NOT need to be stored, we currently store it along with Token B since we need to create and save the record that has Token B as this is then used by MSP, I suppose that can be left out since it is temporary, and quickly overwritten after the response from MSP. Arguably, if something goes wrong, it may be good to see that value in the DB though and since the query is already being made to save the data, I am not sure if it hurts to leave that line there, what do you think?

lydiazcheng: I had the idea to not store token A related things because storing and then remove them very soon are a bit of redundant. But I fully agree your concerns. Let's keep it as it is now.

00_Base/src/services/credentials.service.ts
elliot-sabitov: @lydiazcheng if you are okay with it, perhaps we can do this as a fast follow? I can create a ticket for this on the board, only so that we can merge this in and have that incremental progress.. what do you think?

lydiazcheng: That would be great. Thank you, @elliot-sabitov!

PR #135, opened 2024-08-12
fix: registerCredentialsTokenA not properly returning DTO

elliot-sabitov:
fix: registerCredentialsTokenA not properly returning DTO because clientInformation was not a correct sequelize object making it not possible to $.get({plain: true}) as it does not support nested child properties, so instead getting lazy loading the client information with includes to ensure DTO can be correctly mapped.
chore: adding exception handling for SequelizeUniqueConstraintError to ensure error is printed nicely


00_Base/src/util/middleware/http.exception.handler.ts
```diff
@@ -88,6 +88,12 @@ export class HttpExceptionHandler implements KoaMiddlewareInterface {
               new HttpExceptionBody((err as any).message),
             );
             break;
+          case 'SequelizeUniqueConstraintError':
```
lydiazcheng, line 91: Is this error be thrown when run this line `await clientCpoTenant.save();`? If yes, I think CpoTenant can have multiple client information, i.e., client and server token pairs. How about changing the logic to updating existing Cpo tenant when registering credentials A rather than always create a new one? Then we can avoid to have this error.  Also it more like a internal server error for me rather then bad request. 🤔

elliot-sabitov: Hmm, so in this case what happens is that sequelize ends up with an error since the party_id and country_code combination is already existing. You are correct and this line `await clientCpoTenant.save();` will raise this exception up the chain to be caught in this exception handler and formatted (just basic formatting nothing special) to be included in the response, otherwise it wasn't really saying the error in the response to the request because the exception was actually being caught in the `KoaServer` base:
```
public run(host: string, port: number) {
    this.app.on('error', (err, _ctx) => {
      console.log('Error intercepted by Koa:', err.message);
    });
```
since this message wasnt handled, and the response Koa would send would just be the string - "Internal Server Error", not even in JSON 😅 .

This endpoint will prevent registering already existing credentials though, so currently it is a one time registration for a single CPO, and we would need to create additional layer to support additional CPOs. I think once the endpoints to create CpoTenant are in place, then we should have a nice way to also update a tenant, then in my opinion, it should become much easier to adjust this section to update existing CpoTenant right, because we would then just re-use that same update CPO tenant functionality. This endpoint would simply do the whole flow, while the update CpoTenant endpoint would only update our DB essentially... because we didnt have the multi tenant stuff finalized quite yet, I was holding off on doing this for now, but perhaps we can address this as a fast follow?

lydiazcheng: Thanks for this detailed reply! Understood. Then no other concerns from my side.

I don't approve since I see many new changes files, I am not sure if they are belongs to this PR or just some temp changes. Please let mw know if I should review them.

elliot-sabitov: ah this is that git bug again - when I do the diff between the two branches, it only shows 3 files changed!

thanaParis (approved):
Approved

PR #136, opened 2024-08-13
OCPI Reservation

lydiazcheng:
## Purpose
This PR is to implement OCPI reserveNow and cancelReservation in Command endpoint.

## Changes
1. Implement reserveNow and cancelReservation in the command endpoint
2. Add handlers for them to send the async response to MSP
3. Create a OcpiReservation table so that OCPI is able to match with Core reservation.
4. Related changes in core: https://github.com/citrineos/citrineos-core/pull/222

## Tests
1. Send reserve request Screenshot 2024-08-13 at 12 23 43 PM
2. Reservations are stored in both OCPI and Core Reservation table but reservation is not active Screenshot 2024-08-13 at 12 06 08 PMScreenshot 2024-08-13 at 12 25 45 PM
3. Charger receives the request and sends back Accept response Screenshot 2024-08-13 at 12 26 49 PM
5. Core Reservation is updated to be active Screenshot 2024-08-13 at 12 08 09 PM
6. MSP receives the async response Screenshot 2024-08-13 at 12 28 17 PM
7. Send cancel reservation request Screenshot 2024-08-13 at 12 28 53 PM
8. Charger receives the request and sends back Accept response Screenshot 2024-08-13 at 12 29 27 PM
9. Core Reservation is updated to be not active Screenshot 2024-08-13 at 12 30 13 PM
10. MSP receives the async response (The url contains still "RESERVE_NOW" because I forget to correct it when sending cancel request using postman for testing. But it is not a code issue.) Screenshot 2024-08-13 at 12 31 12 PM


00_Base/src/util/command.executor.ts
```diff
@@ -298,6 +318,126 @@ export class CommandExecutor {
...
+      throw new BadRequestError('Missing evse_uid');
+    }
+    const evse = await this.ocpiEvseEntityRepo.getOcpiEvseByEvseUid(
+      reserveNow.evse_uid!,
+    );
+    if (!evse) {
+      throw new NotFoundError('EVSE not found');
+    }
+    const ocpiLocation = await this.ocpiLocationRepo.readByKey(
+      reserveNow.location_id,
```
yuvenus, line 338: it's misleading, but the ocpi location generated database id isn't actually the "real" ocpi location id 🙈 the core location id is the one used for all ocpi-related operations, while the ocpi location database id is just there to be there.

you should be able to do this:

```
const ocpiLocation = await this.ocpiLocationRepo.getLocationByCoreLocationId(reserveNow.location_id);
```

and it should get you the same result 🤞

00_Base/src/services/commands.service.ts
```diff
@@ -49,22 +59,82 @@ export class CommandsService {
...
+        reserveNow,
+        fromCountryCode,
+        fromPartyId,
+      );
+      return ResponseGenerator.buildGenericSuccessResponse({
+        result: CommandResponseType.ACCEPTED,
+        timeout: this.TIMEOUT,
+      });
+    } catch (e) {
+      if (e instanceof NotFoundError || e instanceof BadRequestError) {
```
yuvenus, line 117: just curious, is there a reason these two errors are combined together to return a not found error? is there just no handling of a `BadRequestError` currently 🤔

03_Modules/Commands/src/module/module.ts
```diff
@@ -49,6 +53,22 @@ export class CommandsModule implements OcpiModule {
         this.logger,
       ),
     );
+
+    Container.set(
```
yuvenus, line 57: building on my comment here https://github.com/citrineos/citrineos-core/pull/222#issuecomment-2289740960

i think if you do what's in the linked comment, you shouldn't need to set here anymore! it'll just inject automatically once the server spins up :eyes:

00_Base/src/model/OcpiReservation.ts
```diff
@@ -0,0 +1,69 @@
...
+  countryCode = 'countryCode',
+  reservationId = 'reservationId',
+  locationId = 'locationId',
+  location = 'location',
+  coreReservationId = 'coreReservationId',
+  coreReservation = 'coreReservation',
+}
+
+@Table
+export class OcpiReservation extends Model {
```
ChrisWeissmann, line 26: Do we need to add EVSEID here as a column? It's an optional in the incoming object, and it seems right now we just throw it away.  As well as `authorization_reference`

00_Base/src/services/commands.service.ts
```diff
@@ -49,22 +59,82 @@ export class CommandsService {
...
+      return ResponseGenerator.buildGenericSuccessResponse({
+        result: CommandResponseType.ACCEPTED,
+        timeout: this.TIMEOUT,
+      });
+    } catch (e) {
+      if (e instanceof NotFoundError) {
+        return ResponseGenerator.buildGenericClientErrorResponse(
+          {
+            result: CommandResponseType.REJECTED,
+            timeout: this.TIMEOUT,
```
ChrisWeissmann, line 82: This is a nitpick and me trying to understand the spec better. But what is the point of a 30s timeout here instead of 0? The rejection is pretty final and the timeout is about the result of the command.

00_Base/src/util/command.executor.ts
```diff
@@ -298,6 +318,126 @@ export class CommandExecutor {
...
+    );
+    if (!ocpiLocation) {
+      throw new NotFoundError('Location not found');
+    }
+
```
ChrisWeissmann, line 343: Do we need to validate that the EVSE belongs to the location that was found?

00_Base/src/util/command.executor.ts
```diff
@@ -360,4 +500,80 @@ export class CommandExecutor {
...
+        type: OcpiTokensMapper.mapOcpiTokenTypeToOcppIdTokenType(
+          reserveNow.token.type,
+        ),
+      },
+      evseId: evse.evseId,
+    } as ReserveNowRequest;
+    const storedCoreReservation =
+      await this.reservationRepo.createOrUpdateReservation(
+        request,
+        evse.stationId,
```
ChrisWeissmann, line 560: It doesn't seem that `createOrUpdateReservation` takes in a stationId separately.

lydiazcheng:
Thanks a lot for the helpful review! @yuvenus @ChrisWeissmann 

I updated the code. But currently I cannot manage to build core and ocpi successfully and there is also a lint dependency issue. I hope the fix that Elliot is working on will help this. I will pull and rebuild after the fixes are merged.

00_Base/src/services/commands.service.ts
lydiazcheng: Good catch! I corrected the code.

03_Modules/Commands/src/module/module.ts
lydiazcheng: Thanks for sharing this!

00_Base/src/model/OcpiReservation.ts
lydiazcheng: Thank you, Christian! I added them in the table. I was struggled with this since these 2 field are not used in other following reserve steps beside reserveNow, `authorization_reference` even not used in reserveNow. But it would be more extendable if we need them in the future.

00_Base/src/util/command.executor.ts
lydiazcheng: I am afraid I did not get this point. Could you please share more for this point?

00_Base/src/util/command.executor.ts
lydiazcheng: This is a good idea. But I am not so clear about the relationship between ocpiEvse and ocpiLocation. Is this possible to do this validation? @yuvenus Do you know more about this?

00_Base/src/services/commands.service.ts
lydiazcheng: I remember @langiole helped share some info about this timeout before but I am cannot remember it clearly.

00_Base/src/util/command.executor.ts
yuvenus: currently ocpiEvse and ocpiLocation don't hold relationships to each other because they are an extension of the ocpp-versions of those objects 😔 

however, there is already logic in `LocationsDatasource` to do the hard work to figure out the relationship. perhaps you can switch from using the repository to using the datasource and you should be able to use the method `getEvse()` which takes in a location id, station id (the first "half" of the evse uid") and the evse id (the second "half" of the evse uid).

there's a helper to split the evse uid into its two different parts, `EXTRACT_STATION_ID` and `EXTRACT_EVSE_ID` both of which take in the evse uid (it's complicated bc evse_uid is not an ocpp field and we have to make one up... 😬)

additionally, `getEvse()` will throw a NotFoundError if
* the location isn't found
* the evse doesn't belong to the location
* the evse isn't found

so i think you can shorten this ocpiEvse and ocpiLocation check into one check to location datasource's `getEvse()` 👀 (big win for the datasources, haha!)

so the code may look more like this:
```
public async executeReserveNow(
    reserveNow: ReserveNow,
    countryCode: string,
    partyId: string,
): Promise<void> {
  // Currently, it rejects the request if the evse_uid is missing
  // When we have the solution to handle the optional evse_uid, we should change this
  if (!reserveNow.evse_uid) {
    throw new BadRequestError('Missing evse_uid');
  }

  // below should already throw as per criteria listed previously
  const evse = await this.locationsDatasource.getEvse(
    reserveNow.location_id,
    EXTRACT_STATION_ID(reserveNow.evse_uid!),
    EXTRACT_EVSE_ID(reserveNow.evse_uid!)
  ); 

... the rest of the method

}
```

00_Base/src/util/command.executor.ts
yuvenus: @ChrisWeissmann i made the same assumption the first time -- this is referencing the core reservation repository, which **does** take in a station id, and not the ocpi reservation repository, which **does not** take in a station id, so it's all good, the arguments are correct!

@lydiazcheng this is what we get for having two similar repositories `reservationRepo` and `ocpiReservationRepo` haha, always get them confused 😄

lydiazcheng: Gotcha. It's indeed confusing. I renamed the repo to `coreReservationRepo`.

00_Base/src/util/command.executor.ts
lydiazcheng: Great! Thank you, Venus! I updated the code.

ChrisWeissmann (approved):

00_Base/src/services/commands.service.ts
thanaParis: I'm going to merge this because I don't think it's an issue either way, but would like to hear if there's a rationale @langiole

PR #137, opened 2024-08-15
chore: updating eslint / tslint setup to newer version

elliot-sabitov:
- chore: updating eslint / tslint setup to newer version and removing no longer supported files but instead using the new eslint.config.js following all recommended settings overriding only the no-explicit-any and no-unused-vars values and setting the ignores
- chore: lint-cleanup, removing lint form sub modules to ensure that there is no conflict, ensuring latest es/tslint config is used


yuvenus (approved):

elliot-sabitov:
this pipeline is currently failing because it is looking for version 1.4.0 which isnt on NPM yet... we can temporarily adjust pipeline to reference the release branch if needed!

00_Base/src/model/OcpiReservation.ts
```diff
@@ -0,0 +1,81 @@
+import { IsNotEmpty, IsString, MaxLength } from 'class-validator';
```
elliot-sabitov, line 1: **Please note** that these reservation related changes are not part of the diff

PR #140, opened 2024-08-16
chore: readding server to provide for a way to run core with ocpi from this repo directly

elliot-sabitov:
chore: correcting path to docker file
chore: removing unnecessary refs
chore: overriding tsconfig.json with tsconfig.docker.json which does not have any refs to ensure that dependencies come from npm
chore: made npm run start-docker-compose convenience command to also --force-recreate to ensure that cached layers are rebuilt
chore: adding docs


yuvenus (approved):
works on my windows machine so far 👀👍

PR #142, opened 2024-08-20
fix: PUT Tariffs endpoint incorrect response body

elliot-sabitov:
fix: PUT Tariffs endpoint to ensure that response body is an OcpiResponse whereas it was only sending the tariff directly
Screenshot 2024-08-20 at 1 34 31 PM


ChrisWeissmann (approved):
Looks good to me. 
Out of curiosity, why the change though?  It's an endpoint for the operator/admin. Why does it have to follow the OCPI wrapped response?

PR #144, opened 2024-08-20
fix: created ContentType enum and reusing in place of application/jso…

elliot-sabitov:
fix: created ContentType enum and reusing in place of application/json string and adjusting exception handlers to set the content type to JSON when handling errors


lydiazcheng (approved):
The changes look good!

PR #146, opened 2024-08-20
chore: removing package lock for now

elliot-sabitov:
chore: removing package lock for now as we investigate a good way to either have a single package lock or support multiple


ChrisWeissmann (approved):

ChrisWeissmann:
@elliot-sabitov can we add it back to the `.gitignore`?

elliot-sabitov:
> .gitignore

added

elliot-sabitov:
> @elliot-sabitov can we add it back to the `.gitignore`?

added

PR #161, opened 2024-11-12
Fix: npm run seed-db & npm run sync-db

elliot-sabitov:
Intended to resolve - https://github.com/citrineos/citrineos/issues/52

- fix: updating sequelize.bridge.config.js to use updated ServerConfig file
- fix: adjusting sequelize.bridge.config.js and db.sync.ts to import reflect-metadata to resolve Reflect not defined error
- fix: installing sequelize-cli because it was not in the dependency tree

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "eslint-config-prettier": "9.1.0",
     "prettier": "3.2.5",
     "typescript": "5.5.4",
-    "typescript-eslint": "7.6.0"
+    "typescript-eslint": "7.6.0",
+    "sequelize-cli": "6.6.2"
   },
   "scripts": {
     "test": "jest --config jest.config.js",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "prettier": "3.2.5",
     "typescript": "5.5.4",
     "typescript-eslint": "7.6.0",
-    "sequelize-cli": "6.6.2"
+    "sequelize-cli": "6.6.2",
+    "reflect-metadata": "0.2.2"
   },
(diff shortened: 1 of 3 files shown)


yuvenus (dismissed):

package.json
```diff
@@ -10,7 +10,8 @@
     "eslint-config-prettier": "9.1.0",
     "prettier": "3.2.5",
     "typescript": "5.5.4",
-    "typescript-eslint": "7.6.0"
+    "typescript-eslint": "7.6.0",
```
yuvenus, line 13: could be worth also adding `"reflect-metadata": "^0.2.2"` for us folks without it already installed on our computers 😄

yuvenus (approved):

package.json
elliot-sabitov: you know - I did add it at first, but then I saw that it was in base, so I thought maybe it would pick it up, and when I ran `npm run fresh` and `npm run install-all` it worked, so I thought it would be okay without it, but I suppose it is better safe the sorry! I added it 👍

lydiazcheng (approved):
Thank for fixing this! It works on my side as well. 👍 
I got 2 small issues but it might be the issue on my own laptop.
1. `ERROR: Error: Error reading "00_Base/src/config/sequalize.bridge.config.js". Error: Error: Cannot find module 'tsconfig-paths/register'` It solves after running `npm install --save-dev tsconfig-paths`
2. This command doesn't work when start OCPI only. How about adding a prerequisite to start both services in`Running database seeders` of readme ?

PR #164, opened 2025-01-16
Update device model queries to include relevant Variable Attribute type

### LINKED ISSUE #98: (OCPI) Retrieving device model should also query based on type "Actual"
**Which repository?**
citrineos-ocpi

**Describe the bug**
When retrieving device model information from the database to populate OCPI data about a charging station, evse, and connector, where appropriate, we should also be querying based on `VariableAttributes.type === "Actual"`, in addition to the rest of the query.

A sample query today that will fail if there are multiple VariableAttributes fulfilling the query, even though the `type` is different:

```
select va."value" 
        from "VariableAttributes" va 
          left join "Variables" v on va."variableId" = v."id" 
          left join "Components" c on va."componentId" = c."id"
          left join "Evses" e on c."evseDatabaseId" = e."databaseId" 
        where va."stationId" = '${stationId}' and e."id" = ${evseComponentId} and e."connectorId" = ${connectorId} and c."name" = 'Connector' and v."name" = 'AvailabilityState'
```

The above query should also include the final condition `and v."type" = 'Actual'` to prevent a "duplicate" errors.

**To Reproduce**
1. Run citrineos-ocpi
2. Have an EVSE with device model that has two entries for connector AvailabilityState, one with type "Actual" and one with type "Target"
3. Retrieve a location, evse, or connector via OCPI API

**Expected behavior**
You should be able to retrieve a location, evse, or connector via OCPI, even if there are device model entries with differing types.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - citrineos-core: rc-1.6.0

yuvenus:
Resolves issue https://github.com/citrineos/citrineos/issues/98 - this PR adds an additional condition to the SQL queries that retrieve relevant device model information to also include the relevant Variable Attribute type.

### DIFF
--- a/00_Base/src/model/variableattributes/EvseVariableAttributes.ts
+++ b/00_Base/src/model/variableattributes/EvseVariableAttributes.ts
@@ -52,11 +52,13 @@ export const CONSTRUCT_EVSE_VARIABLE_ATTRIBUTES_QUERY = (
   from 
     coalesce(
       (
-        select string_agg(distinct e."connectorId"::text, ',') from "VariableAttributes" va 
-        left join "Variables" v on va."variableId" = v."id" 
-        left join "Components" c on va."componentId" = c."id"
-        left join "Evses" e on c."evseDatabaseId" = e."databaseId"
-        where va."stationId" = '${stationId}' and e."connectorId" is not null
+        select string_agg(distinct e."connectorId"::text, ',') 
+        from "VariableAttributes" va 
+          left join "Variables" v on va."variableId" = v."id" 
+          left join "Components" c on va."componentId" = c."id"
+          left join "Evses" e on c."evseDatabaseId" = e."databaseId"
+        where va."stationId" = '${stationId}' and 
+          e."connectorId" is not null
       ), ''
     ) as connector_ids_string,
     coalesce(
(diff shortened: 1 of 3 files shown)


00_Base/src/model/variableattributes/EvseVariableAttributes.ts
```diff
@@ -52,11 +52,13 @@ export const CONSTRUCT_EVSE_VARIABLE_ATTRIBUTES_QUERY = (
...
-        left join "Variables" v on va."variableId" = v."id" 
-        left join "Components" c on va."componentId" = c."id"
-        left join "Evses" e on c."evseDatabaseId" = e."databaseId"
-        where va."stationId" = '${stationId}' and e."connectorId" is not null
+        select string_agg(distinct e."connectorId"::text, ',') 
+        from "VariableAttributes" va 
+          left join "Variables" v on va."variableId" = v."id" 
+          left join "Components" c on va."componentId" = c."id"
+          left join "Evses" e on c."evseDatabaseId" = e."databaseId"
+        where va."stationId" = '${stationId}' and
```
lydiazcheng, line 60: One question: why does this query as well as query for `evse_ids_string` don't need to add `va."type" = 'Actual' and`?

lydiazcheng (approved):

00_Base/src/model/variableattributes/EvseVariableAttributes.ts
yuvenus: since it's the unique set of the evse ids, we don't necessarily have to target the type since if we pull let's say variable attributes of type Actual and Target with the same evse, they'll collapse into one evse in the set 👍

PR #166, opened 2025-01-28
Updating ocpi to use new OCPP enums

eddygrebt:
Ensuring the OCPI will run with citrineOS core 1.6.0.


00_Base/src/services/TransactionFilterService.ts
```diff
@@ -56,18 +56,13 @@ export class TransactionFilterService implements ITransactionDatasource {
...
-        >,
-      ),
+      Transaction.count(countQueryOptions),
     ]);
 
     const result: PaginatedResult<Transaction> =
       new PaginatedResult<Transaction>();
     result.data = transactions;
-    result.total = total;
+    result.total = total[0].count;
```
lydiazcheng, line 65: If the omit group part is removed, do we need to get a sum of the `total` array like below? or total[0] is always the actual total number?
```suggestion
    if (total.length === 1) {
      result.total = total[0].count;
    } else {
      result.total = total.reduce((sum, group) => (sum + group.count), 0);
    }
```

03_Modules/Cdrs/src/module/handlers.ts
```diff
@@ -62,11 +63,7 @@ export class CdrsOcppHandlers extends AbstractModule {
     const timer = new Timer();
     this._logger.info('Initializing...');
 
-    if (!deasyncPromise(this._initHandler(this._requests, this._responses))) {
-      throw new Error(
-        'Could not initialize module due to failure in handler initialization.',
-      );
-    }
+    this.initHandlers();
```
lydiazcheng, line 66: How about init handlers centrally in the `Server/src/index.ts` like [here](https://github.com/citrineos/citrineos-core/blob/9432540f35b5efe0369e2f02dc1fb1eb032b88c0/Server/src/index.ts#L461) rather than in each `handlers`. I remember this is what @thanaParis suggested in the daily. Then we also don't need `00_Base/src/util/Timer.ts` in this case. (If I misunderstand or missed anything, pls feel free to let me know)

lydiazcheng (commented):
The change looks good. Only 2 questions from my side.

PR #171, opened 2025-06-25
TC-21: Migrate service datasource call to graphql

Adex9ja:


00_Base/src/services/TariffsService.ts
```diff
@@ -5,23 +5,31 @@ import { OcpiTariff, TariffKey } from '../model/OcpiTariff';
...
+    private readonly ocpiGraphqlClient: OcpiGraphqlClient,
   ) {}
 
   async getTariffByKey(key: TariffKey): Promise<TariffDTO | undefined> {
-    return this.tariffsDatasource.getTariffByKey(key);
+    const result = await this.ocpiGraphqlClient.request<GetTariffByKeyQuery>(GET_TARIFF_BY_KEY_QUERY, key);
+    return result.Tariffs?.[0] as unknown as TariffDTO | undefined;
   }
 
   async getTariffsForOcpiTariffs(ocpiTariffs: OcpiTariff[]) {
```
thanaParis, line 28: Why do we still need this function? There are no OcpiTariffs to map anymore, I think you should delete it.

thanaParis (commented):
This is looking good!

thanaParis (approved):
approved

PR #183, opened 2025-07-15
TC-51: graphql api for session pull requests

Adex9ja:
## Summary
This PR implements a GraphQL query for retrieving Transactions using cpoCountryCode, cpoPartyId, with optional filtering on clientCountryCode and clientPartyId only when provided and mapping them to session.

## Changes Made
- Introduced conditional logic for clientCountryCode and clientPartyId in GraphQL where clause
- Updated GET_TRANSACTIONS_QUERY and related service logic to support optional filters
- Refactored relevant mappers and DTOs to align with updated query structure
- Ensured backward compatibility by skipping null filters
- Added endedOnly flag to return filtered sessions with energy usage

CitrineOS Changes: https://github.com/citrineos/citrineos-core/pull/444


00_Base/src/services/CdrsService.ts
```diff
@@ -21,19 +25,26 @@ export class CdrsService {
...
       dateTo: dateTo ? dateTo.toISOString() : undefined,
       offset,
       limit,
     };
-    const result = await this.ocpiGraphqlClient.request<any>(GET_TRANSACTIONS_QUERY, variables);
+    const result = await this.ocpiGraphqlClient.request<GetTransactionsQuery>(
+      GET_TRANSACTIONS_QUERY,
+      variables,
+    );
+    const mappedCdr = await this.cdrMapper.mapTransactionsToCdrs(
```
thanaParis, line 41: This method/class uses repositories and cannot be used without refactoring.

00_Base/src/services/SessionsService.ts
```diff
@@ -39,14 +45,22 @@ export class SessionsService {
       queryOptions,
     );
 
+    let mappedSessions = await this.sessionMapper.mapTransactionsToSessions(
```
thanaParis, line 48: This method/class uses repositories and cannot be used without refactoring.

00_Base/src/services/CdrsService.ts
```diff
@@ -3,11 +3,15 @@ import { DEFAULT_LIMIT, DEFAULT_OFFSET } from '../model/PaginatedResponse';
 import { PaginatedCdrResponse } from '../model/Cdr';
 import { OcpiGraphqlClient } from '../graphql/OcpiGraphqlClient';
 import { GET_TRANSACTIONS_QUERY } from '../graphql/queries/transaction.queries';
+import { GetTransactionsQuery } from '../graphql/types/graphql';
+import { CdrMapper } from '../mapper/CdrMapper';
+import { Transaction } from '@citrineos/data';
```
thanaParis, line 8: You should not be using this class anywhere in this pr. Please use the appropriate interface (ITransactionDto)

thanaParis (changes requested):
Changes needed.

thanaParis (changes requested):
one small change, otherwise looks good!

PR #185, opened 2025-07-17
refactor: move AsyncJobStatus models to @citrineos/data package

deye9:
- Remove local AsyncJobStatus, AsyncJobAction, and AsyncJobRequest models
- Update imports to use consolidated types from @citrineos/data
- Delete AsyncJobStatusRepository as it's now handled centrally
- Consolidate async job functionality to reduce code duplication

This change centralizes async job status management in the shared data package,
improving maintainability and consistency across CitrineOS modules.


thanaParis (changes requested):
changes needed.

00_Base/src/util/OcpiSequelizeInstance.ts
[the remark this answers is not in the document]
deye9: @thanaParis 

I can see that OcpiSequelizeInstance is heavily used throughout the codebase. If it's being deleted, this represents a major architectural change. The main usages are in:

index.ts - Core integration
db.sync.ts - Database synchronization
emsp.server.ts - Mock server

**My ask:**
What's replacing OcpiSequelizeInstance?
Is the project moving to use @citrineos/data package's Sequelize instance directly?
Is there a new database abstraction layer?
Does this affect the GraphQL migration we just completed?


Is the deletion part of what I should handle in this ticket?

00_Base/src/services/TokensAdminService.ts
```diff
@@ -91,14 +94,14 @@ export class TokensAdminService {
...
       const params = buildPaginatedOcpiParams(
-        asyncJobStatus.mspCountryCode,
-        asyncJobStatus.mspPartyId,
-        asyncJobStatus.cpoCountryCode,
-        asyncJobStatus.cpoPartyId,
-        asyncJobStatus.paginationParams.offset,
-        asyncJobStatus.paginationParams.limit,
-        asyncJobStatus.paginationParams.dateFrom,
-        asyncJobStatus.paginationParams.dateTo,
+        asyncJobStatus.tenantPartner?.mspCountryCode ?? '',
```
thanaParis, line 97: I'm confused here. why does tenantPartner have 'mspCountryCode' and 'cpoCountryCode'? In CitrineosCore, tenantPartner has 'countryCod' (which is the msp countryCode) and _tenant_.countryCode (not tenantPartner) is the cpoCountryCode. This mapping is needlessly confusing and should follow the way the objects are defined in core.

00_Base/src/types/asyncJob.types.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2023 S44, LLC
+// Copyright Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache 2.0
+
+// Temporary local definitions until @citrineos/data exports are fixed
```
thanaParis, line 6: You should be using Dto interface exports from citrineos/base, not citrineos/data. These shouldn't need to be here.

thanaParis (changes requested):
some changes needed

00_Base/src/services/TokensAdminService.ts
```diff
@@ -18,11 +16,21 @@ import {
...
+// Import AsyncJob types from the base module exports which re-export from @citrineos/data
+import {
+  AsyncJobRequest,
+  AsyncJobStatus,
+  AsyncJobName,
+  SequelizeAsyncJobStatusRepository,
+} from '../index';
+
+// Type alias for the repository interface expected by the service
+type AsyncJobStatusRepository = SequelizeAsyncJobStatusRepository;
```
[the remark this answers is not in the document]
deye9: Fixed

00_Base/src/types/asyncJob.types.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// AsyncJob DTOs following @citrineos/base patterns
+export interface AsyncJobPaginatedParams {
+  offset?: number;
+  limit?: number;
+  dateFrom?: Date;
+  dateTo?: Date;
+}
+
+export interface AsyncJobStatusDTO extends IBaseDto {
```
thanaParis, line 26: I'm confused by this type. Since AsyncJobs are now in citrineos-core's database, they should have a Dto class there. This class is also being used for the response type for an API, which I think should be a different interface.
There should be:
- A type in citrineos-core's 00_Base alongside the other IDto interfaces that extends IBaseDto
- A type here in citrineos-ocpi used as the response type by the API

thanaParis (changes requested):
one change needed, I think there are types in this PR which should be in citrineos-core

PR #193, opened 2025-08-11
TC-95: Seed default Tenant and Partner

Adex9ja:
Seeder files for adding default Tenant and Tenant Partner


thanaParis (approved):
approved

PR #229, opened 2026-03-13
Fix proper OCPI 2.2.1 implementation as sender

florianne1212:
I found a few issues for OCPI 2.2.1 implementation:
1 - Token B must be set before making outgoing version requests
According to OCPI 2.2.1, Token B (received from the partner during registration) must be used to authenticate outgoing requests to GET /versions and GET /version-details. In the original implementation, credentials was being set on partnerProfileOCPI after Token C (called token B in the code) was generated, meaning it was not available on the profile when the version requests were made. This was inconsistent with the OCPI spec and with the role of credentials in this codebase (see Fix 2).

Fix: Token A is now stored in partnerProfileOCPI.credentials before getVersionDetails() is called, ensuring it is available as the outgoing auth token for GET /versions and GET /version-details.

2 - serverCredentials.token was used for outgoing request authentication
From my understanding of the codebase it follows this convention:
serverCredentials.token is the token we issue to our partner to authenticate their incoming requests to us
credentials.token is the token our partner issued to us, used to authenticate our outgoing requests to them

BaseClientApi.getHeaders() was using partnerProfile.serverCredentials.token to set the Authorization header on all outgoing HTTP calls, which is the wrong token — it was sending our own inbound token instead of the partner's token.

Fix: Changed getHeaders() to use partnerProfile.credentials?.token.

3 - Wrong schema used in VersionsClientApi.getVersions()
getVersions() was validating the CPO's response against VersionDetailsResponseDTOSchema, which expects data to be an This caused a runtime Zod validation crash:
 Expected object, received array at path data
Fix: Changed the schema passed to request() from VersionDetailsResponseDTOSchema to VersionListResponseDTOSchema, which correctly expects z.array(VersionDTOSchema). 

[](url)
Screenshot_20260313_104212

### DIFF
--- a/00_Base/src/trigger/BaseClientApi.ts
+++ b/00_Base/src/trigger/BaseClientApi.ts
@@ -92,7 +92,7 @@ export abstract class BaseClientApi {
     const headers: IHeaders = {};
     headers[OcpiHttpHeader.XRequestId] = uuidv4();
     headers[OcpiHttpHeader.XCorrelationId] = uuidv4();
-    const token = partnerProfile.serverCredentials.token;
+    const token = partnerProfile.credentials?.token;
     if (!token) {
       throw new MissingRequiredParamException(
         'token',
--- a/00_Base/src/services/CredentialsService.ts
+++ b/00_Base/src/services/CredentialsService.ts
@@ -100,6 +100,10 @@ export class CredentialsService {
         `TenantPartner expects ${partner.partnerProfileOCPI!.version.version}, received ${versionNumber}`,
       );
     }
+    partner.partnerProfileOCPI!.credentials = {
+      versionsUrl: credentials.url,
+      token: credentials.token,
+    };
     const tenantPartner = await this.getVersionDetails(
       partner,
       credentials.url,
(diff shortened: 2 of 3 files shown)


00_Base/src/trigger/BaseClientApi.ts
```diff
@@ -92,7 +92,7 @@ export abstract class BaseClientApi {
     const headers: IHeaders = {};
     headers[OcpiHttpHeader.XRequestId] = uuidv4();
     headers[OcpiHttpHeader.XCorrelationId] = uuidv4();
-    const token = partnerProfile.serverCredentials.token;
+    const token = partnerProfile.credentials?.token;
```
lydiazcheng, line 95: Hi @florianne1212 , thanks for the changes.
The server token is generated by the server (ocpi) and shared with clients, e.g., it is token B when ocpi is the sender in the registration process. In this case, serverCredentials.token should be set in the header and sent to client for authentication.

lydiazcheng (commented):
The PR looks good. Only left one comment. Could you pls update the destination branch to `next` which we use as the release candidate? Thank you.

thanaParis (approved):

